PR #67, opened 2022-10-04
BarChart and LineChart

LeeAlephium:
ApexCharts is not available for ReactNative.

React-Native-Svg, d3-shape and d3-scale were used instead.

They were first implemented here:

* https://jsfiddle.net/dukhewc1/12/
* https://jsfiddle.net/048rsqt2/

Then I've implemented the BarChart. LineChart is tomorrow.


nop33:
Superseded by #70. Closing.

PR #88, opened 2022-10-19
Pull to refresh & dark theme fixes

nop33:


src/store/addressesSlice.ts
```diff
@@ -105,18 +104,9 @@ export const fetchAddressesDataNextPage = createAsyncThunk(
...
-      let newTransactions = [] as Transaction[]
-
-      if (!allPagesLoaded) {
-        nextPage += 1
-        console.log(`⬇️ Fetching page ${nextPage} of address confirmed transactions: `, addressHash)
-        const { data: transactions } = await client.explorerClient.getAddressTransactions(addressHash, nextPage)
-        newTransactions = transactions
-      }
+      const page = 1
+      console.log(`⬇️ Fetching page ${page} of address confirmed transactions: `, addressHash)
```
mvaivre, line 108: Should we differentiate the log message from the one from `fetchAddressesDataNextPage`?
We fetch more than just the transactions here.
Also, if `page` is a constant, let's just remove it?

src/utils/transactions.ts
```diff
@@ -0,0 +1,35 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { Transaction } from '@alephium/sdk/api/explorer'
+
+export const getNewTransactions = (
```
mvaivre, line 21: `extractNewTransactions`? Just to differentiate it from an endpoint call.

mvaivre (commented):
👌

src/utils/transactions.ts
nop33: Yep, sounds appropriate!

src/store/addressesSlice.ts
nop33: Yes, I'll rework it a bit, thanks!

src/utils/transactions.ts
nop33: I included the fix in the next PR, I hope that's fine 😇 efd336e

src/store/addressesSlice.ts
nop33: Same for this fix :) a0a2b3f

src/utils/transactions.ts
mvaivre: Perfect :)

mvaivre (approved):

Issue #74, opened 2022-10-18
Handle UTXO consolidation edge-case when sending ALPH


nop33:
Similar to how the desktop wallet feature where if the node replies with an error that it cannot build the tx due to too many UTXOs, the mobile wallet needs to handle UTXO consolidation.

Here's the relevant desktop wallet code:
- Detecting error from API: https://github.com/alephium/desktop-wallet/blob/master/src/modals/SendModal/index.tsx#L104-L132
- Building consolidation transaction: https://github.com/alephium/desktop-wallet/blob/master/src/modals/SendModal/index.tsx#L145-L160
- Displaying consolidation modal: https://github.com/alephium/desktop-wallet/blob/master/src/modals/SendModal/index.tsx#L247-L255

PR #90, opened 2022-10-25
Rework tx sending workflow, tx consolidation, and tx display

### LINKED ISSUE #89: (Production preview) wallet crashes when brought from bg to fg
You can download an APK here to test this out: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/48e7075a-d986-41f6-8b69-7f092ad80c01

nop33:
I am opening this PR against the `refresh` branch which has an open PR: #88

1. It improves the way pending txs are handled. Instead of merging "dummy" data into the confirmed transactions array in the app state, it stores them separately. This makes the code much cleaner, imo.
2. It handles the UTXO consolidation corner case, when the address the user tries to send funds from has too many UTXOs. Fixes https://github.com/alephium/mobile-wallet/issues/74
3. It updates the display of transactions to look sleek af (and aligned with how they look in all other apps of ours).
4. It simplifies A LOT the process of restoring the navigation state
5. It fixes #89 

Here's a preview (in shity quality due to GH filesize limit)

https://user-images.githubusercontent.com/1579899/197814106-f629b16d-13f3-4841-b70b-b30c53e3f49a.mov


mvaivre:
That's magnificent.

src/screens/SendScreen.tsx
```diff
@@ -0,0 +1,361 @@
...
+  const [amountString, setAmountString] = useState('')
+  const [amount, setAmount] = useState(BigInt(0))
+  const [fees, setFees] = useState<bigint>(BigInt(0))
+  const [unsignedTxId, setUnsignedTxId] = useState('')
+  const [unsignedTransaction, setUnsignedTransaction] = useState('')
+  const [gasAmount, setGasAmount] = useState('')
+  const [gasPriceString, setGasPriceString] = useState('')
+  const [gasPrice, setGasPrice] = useState<bigint>()
+  const [isLoading, setIsLoading] = useState(false)
+  const [isConsolidateUTXOsModalVisible, setIsConsolidateUTXOsModalVisible] = useState(false)
```
mvaivre, line 83: That's a lot of states 🙃
I wonder if we should maybe start considering using something like https://react-hook-form.com/ which would greatly simplify local form state management and validation.

nop33: I'll have a look into it!

src/utils/transactions.ts
```diff
@@ -18,18 +18,21 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-    } else {
-      if (existingTransactions[targetTxIndex].blockHash === '') {
-        existingTransactions.splice(targetTxIndex, 1, incomingTx)
-      }
-    }
-    return newTxs
-  }, [])
+  incomingTransactions.filter((newTx) => !existingTransactions.some((existingTx) => existingTx.hash === newTx.hash))
+
+export const getRemainingPendingTransactions = (
```
mvaivre: Same comment as before, `extractRemainingPendingTransactions` to avoid confusion with endpoint call?

nop33: haha, already done!

mvaivre: Ah you did it already :D

src/screens/SendScreen.tsx
nop33: Hm, looking at the [integration process of React Hook Form into React Native](https://echobind.com/post/react-hook-form-for-react-native) and the changes that are needed to our codebase, I am not so sure it's worth going for it for one component that has a lot of state. WDYT? I could try reducing the use of `useState` by grouping things together...

nop33: Or maybe I am wrong, wait: https://react-hook-form.com/get-started#reactnative

src/screens/SendScreen.tsx
```diff
@@ -68,113 +81,119 @@ const SendScreen = ({
...
-      setGasPrice(convertAlphToSet(gasPriceString))
-    } catch {
-      setGasPrice(undefined)
-    }
-  }, [gasPriceString])
+  const isFormValid = isEmpty(errors)
+  const totalAmount = amount + fees
+
+  const validateIsAddressValid = (value: string) => isAddressValid(value) || 'This address is not valid'
+  const validateIsNumericStringValid = (value: string) => isNumericStringValid(value) || 'We need a number here!'
```
mvaivre, line 116: Should we start to extract generic validation method out of this component?

Also, let's maybe replace the error by something like `A number is expected.` :)

mvaivre:
One tiny comment. Otherwise, the code looks great to me! Well done!!

src/screens/SendScreen.tsx
nop33: done!

src/navigation/RootStackNavigation.tsx
```diff
@@ -84,26 +78,12 @@ const RootStackNavigation = () => {
...
-        // Means this is a regular login, not coming back from being inactive
-      } else if (isAuthenticated && isAppBackgroundedAcknowledged && !isResetFromBeingInactive) {
-        rootStackNavigationRef.resetRoot({ index: 0, routes: [{ name: 'InWalletScreen' }] })
-      }
-    },
+  useEffect(() => {
+    if (isAuthenticated) {
+      rootStackNavigationRef.resetRoot(lastNavigationState ?? { index: 0, routes: [{ name: 'InWalletScreen' }] })
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
mvaivre, line 85: Is this needed?

mvaivre (approved):
👏

PR #93, opened 2022-10-27
Implement new seed import screen

### LINKED ISSUE #73: Implement new mnemonic import screen
There's also the idea of allowing the import of mnemonic by scanning a QR code that could be displayed in the desktop wallet.

nop33:
Closes #73

Check this out! 🚀 

I know you said "No paste"... but... 😇  Let me know if you strongly disagree with the implementation and I'll remove it.

https://user-images.githubusercontent.com/1579899/198331240-7e372400-2c37-4663-afff-38409f308057.mov


nop33:
As requested on Slack, I removed the processing of pasted mnemonic. I would argue that we should not disable the OS default pasting functionality because this might cause frustration to users. It would require disabling the input field's context menu, which can be enabled either by long press or by double tap. However, this would disable all other context menu options, which might frustrate users (I def get frustrated when a website changes OS-specific things like scrolling etc). I propose that we instead not process strings that include spaces and display a message like in the video below:

https://user-images.githubusercontent.com/1579899/198970015-911e8b8b-8a39-403b-ac2f-dba44e7f30fd.mp4

mvaivre:
100% agreed. That looks perfect.

src/screens/new-wallet/ImportWalletSeedScreen.tsx
```diff
@@ -17,43 +17,76 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-      secretPhrase
-        .trim()
-        .split(' ')
-        .filter((word) => word.length > 2)
+    setPossibleMatches(
+      typedInput.length <= 2
+        ? []
+        : allowedWords.current.filter((allowedWord) => allowedWord.startsWith(typedInput.toLowerCase().trim()))
+    )
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mvaivre, line 64: I've seen you doing this twice, which in my mind is not a good practice :) Have you read more on the subject?
Is it better to ignore the warning or to compare the previous and current state before executing this effect?

nop33: This is leftovers from my previous code, I should have removed the line!

mvaivre (approved):
Yay!

PR #95, opened 2022-10-31
Add "use max" amount & check for authentication requirement before sending

nop33:


src/screens/SendScreen.tsx
```diff
@@ -122,10 +127,7 @@ const SendScreen = ({
     convertAlphToSet(value) >= MINIMAL_GAS_PRICE ||
     `Gas price must be at least ${formatAmountForDisplay(MINIMAL_GAS_PRICE, true)}`
 
-  useEffect(() => {
-    if (txStep === 'send') setTxStep('build')
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fromAddress, toAddressHash, amountInAlph, gasAmount, gasPriceInAlph])
```
nop33, line 128: @mvaivre Thanks for making me more conscious about my use of disabling the eslint rule. I figured I can simplify the code since `setState` [does not cause a re-render when the value didn't change](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update). Now, there's no need to disable the rule.

src/components/ConfirmWithAuthModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+          )
+        } else {
+          console.error(e)
+        }
+      }
+    }
+
+    getWallet()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
nop33, line 79: I want this effect to only run exactly 1 time. If I enable the exhaustive deps rule, I would need to add `onCancel` and `onConfirm` in the dependency list. Those 2 callbacks are defined here:

```ts
// src/screens/SendScreen.tsx
<ConfirmWithAuthModal onCancel={() => setIsAuthenticationModalVisible(false)} onConfirm={sendTransaction} />
```
https://github.com/alephium/mobile-wallet/blob/send-screen-additions/src/screens/SendScreen.tsx#L449

Since the `onCancel` callback gets re-defined every time `SendScreen` gets re-rendered, my effect would run **more than exactly once**, which is not desired. To avoid this I would have to extract the callback definition from the render function and wrap it around a `useCallback`. That would work for `onCancel` (although I find it overengineering for such a simple callback).

But it wouldn't for `onConfirm` because in the dependency list of the `sendTransaction`'s useCallback there is `fromAddress` which changes when we send a transaction (due to the pending tx added to it).  We could of course update the code to remove `fromAddress` from the depedency list, but that would require changing our `Client` class functions to accept `publicKey`, `privateKey`, and `hash` instead of just accepting the `Address` object.

All in all, even though I understand the need for this rule and even though it has saved my ass many times, I don't think it should be a hard rule.

I wonder what your opinion is on this.

nop33: This is basically the diff. As you can see, the `onCancel` had to be extracted and wrapped in a `useCallback` and the dependency array of `sendTransaction` has to be tightened up (removing `fromAddress` and adding `fromAddress?.hash` and `fromAddress?.privateKey` instead, fields that do not change when adding a pending transaction to `fromAddress`.

Even though the code doesn't feel as clean now, I tend to feel that the code is more secure and more performant, since it causes fewer re-renders.

I look forward to your opinion on this.

<details><summary>Click to see it</summary>

```diff
diff --git a/src/api/client.ts b/src/api/client.ts
index 1c6e995..e0b3b32 100644
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -38,8 +38,8 @@ export class Client {
     await this.cliqueClient.init(isMultiNodesClique)
   }
 
-  async buildSweepTransactions(address: Address, toHash: AddressHash) {
-    const { data } = await this.cliqueClient.transactionConsolidateUTXOs(address.publicKey, address.hash, toHash)
+  async buildSweepTransactions(fromHash: AddressHash, fromPublicKey: string, toHash: AddressHash) {
+    const { data } = await this.cliqueClient.transactionConsolidateUTXOs(fromPublicKey, fromHash, toHash)
     const fees = data.unsignedTxs.reduce((acc, tx) => acc + BigInt(tx.gasPrice) * BigInt(tx.gasAmount), BigInt(0))
 
     return {
@@ -48,9 +48,9 @@ export class Client {
     }
   }
 
-  async signAndSendTransaction(fromAddress: Address, txId: string, unsignedTx: string) {
-    const signature = this.cliqueClient.transactionSign(txId, fromAddress.privateKey)
-    const { data } = await this.cliqueClient.transactionSend(fromAddress.hash, unsignedTx, signature)
+  async signAndSendTransaction(fromHash: AddressHash, fromPrivateKey: string, txId: string, unsignedTx: string) {
+    const signature = this.cliqueClient.transactionSign(txId, fromPrivateKey)
+    const { data } = await this.cliqueClient.transactionSend(fromHash, unsignedTx, signature)
 
     return data
   }
diff --git a/src/components/ConfirmWithAuthModal.tsx b/src/components/ConfirmWithAuthModal.tsx
index 020ebbb..be6c025 100644
--- a/src/components/ConfirmWithAuthModal.tsx
+++ b/src/components/ConfirmWithAuthModal.tsx
@@ -75,8 +75,7 @@ const ConfirmWithAuthModal = ({ onConfirm, onCancel }: ConfirmWithAuthModalProps
     }
 
     getWallet()
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [onCancel, onConfirm])
 
   useEffect(() => {
     if (!requiresPin || !pinCode || !activeWallet) return
diff --git a/src/screens/SendScreen.tsx b/src/screens/SendScreen.tsx
index cd00dec..76b73fd 100644
--- a/src/screens/SendScreen.tsx
+++ b/src/screens/SendScreen.tsx
@@ -136,7 +136,11 @@ const SendScreen = ({
     setIsLoading(true)
 
     try {
-      const { unsignedTxs, fees } = await client.buildSweepTransactions(fromAddress, fromAddress.hash)
+      const { unsignedTxs, fees } = await client.buildSweepTransactions(
+        fromAddress.hash,
+        fromAddress.publicKey,
+        fromAddress.hash
+      )
       setSweepUnsignedTxs(unsignedTxs)
       setFees(fees)
     } catch (e) {
@@ -148,7 +152,7 @@ const SendScreen = ({
 
   const buildTransaction = useCallback(
     async (formData: FormData) => {
-      if (!fromAddress || !isFormValid) return
+      if (!fromAddress?.hash || !isFormValid) return
 
       setIsLoading(true)
       const amountInSet = convertAlphToSet(formData.amountInAlph)
@@ -160,7 +164,11 @@ const SendScreen = ({
       const gasPriceInSet = formData.gasPriceInAlph ? convertAlphToSet(formData.gasPriceInAlph) : ''
       try {
         if (isSweep) {
-          const { unsignedTxs, fees } = await client.buildSweepTransactions(fromAddress, formData.toAddressHash)
+          const { unsignedTxs, fees } = await client.buildSweepTransactions(
+            fromAddress.hash,
+            fromAddress.publicKey,
+            formData.toAddressHash
+          )
           setSweepUnsignedTxs(unsignedTxs)
           setFees(fees)
         } else {
@@ -194,18 +202,24 @@ const SendScreen = ({
         setIsLoading(false)
       }
     },
-    [buildConsolidationTransactions, fromAddress, isFormValid]
+    [
+      buildConsolidationTransactions,
+      fromAddress?.hash,
+      fromAddress?.networkData.availableBalance,
+      fromAddress?.publicKey,
+      isFormValid
+    ]
   )
 
   const sendTransaction = useCallback(async () => {
-    if (!fromAddress) return
+    if (!fromAddress?.hash) return
 
     setIsLoading(true)
 
     try {
       if (isSweeping) {
         for (const { txId, unsignedTx } of sweepUnsignedTxs) {
-          client.signAndSendTransaction(fromAddress, txId, unsignedTx)
+          client.signAndSendTransaction(fromAddress.hash, fromAddress.privateKey, txId, unsignedTx)
 
           dispatch(
             addPendingTransactionToAddress({
@@ -219,7 +233,7 @@ const SendScreen = ({
           )
         }
       } else {
-        client.signAndSendTransaction(fromAddress, unsignedTxId, unsignedTransaction)
+        client.signAndSendTransaction(fromAddress.hash, fromAddress.privateKey, unsignedTxId, unsignedTransaction)
 
         dispatch(
           addPendingTransactionToAddress({
@@ -241,7 +255,8 @@ const SendScreen = ({
     amount,
     consolidationRequired,
     dispatch,
-    fromAddress,
+    fromAddress?.hash,
+    fromAddress?.privateKey,
     fromAddressHash,
     isSweeping,
     navigation,
@@ -265,6 +280,8 @@ const SendScreen = ({
     setValue('amountInAlph', convertSetToAlph(BigInt(fromAddress.networkData.availableBalance)))
   }, [fromAddress, setValue])
 
+  const handleAuthCancel = useCallback(() => setIsAuthenticationModalVisible(false), [])
+
   return (
     <Screen>
       <ScrollView>
@@ -446,7 +463,7 @@ const SendScreen = ({
           </ConsolidationModalContent>
         </ModalWithBackdrop>
         {isAuthenticationModalVisible && (
-          <ConfirmWithAuthModal onCancel={() => setIsAuthenticationModalVisible(false)} onConfirm={sendTransaction} />
+          <ConfirmWithAuthModal onCancel={handleAuthCancel} onConfirm={sendTransaction} />
         )}
       </ScrollView>
     </Screen>

```

</details>

nop33: I decided to add a commit that fixes this, I can always revert it if needed.

mvaivre: I've been wondering the same many times. I've encountered this problem many times, and most of those times I was able to use either `useCallback`, `useMemo` or code restructuration to avoid unwanted executions...
But sometimes it's indeed a huge pain in the ass...

So I'd say:
1) We try to make it work with one of the method cited above
2) But if we see that this is forcing us to overengineer too many things, we ignore the rule (if we're sure that it won't produce any unwanted side-effects).

src/screens/SendScreen.tsx
mvaivre: Awesome 🎉

src/components/ConfirmWithAuthModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+        const storedActiveWallet = await getStoredActiveWallet()
+
+        if (storedActiveWallet) {
+          if (storedActiveWallet.authType === 'biometrics') {
+            onConfirm()
+          } else if (storedActiveWallet.authType === 'pin') {
+            setActiveWalletEncryptedMnemonic(storedActiveWallet.mnemonic)
+          }
+        }
+      } catch (e: unknown) {
```
mvaivre, line 57: Typing errors strictly is often complicated (I stumbled upon this issue many times...).
Here, I feel like we can do better, as the expo packages are all written in Typescript, including `expo-local-authentication`. 
After a quick look, I see that this package exports potentially interesting types:
```ts
export { LocalAuthenticationOptions, AuthenticationType, LocalAuthenticationResult, SecurityLevel };
```

It may be worth taking a quickly look to see if this can be improved. If too complicated, let's not bother too much and just keep it as it is right now!

nop33: I will have a look at it! However, the error I am trying to catch here is an error from `expo-secure-store` and specifically an error from `SecureStore.getItemAsync`. `expo-local-authentication` is not used here.

The docs of `expo-secure-store` for `getItemAsync` say:

> The promise will reject if an error occurred while retrieving the value.

https://docs.expo.dev/versions/latest/sdk/securestore/

Looking at the source code, however, I can see that they do have some error codes. They are just not exported.

image

nop33: I created this issue https://github.com/alephium/mobile-wallet/issues/98 and I'll merge this PR, if that's ok :)

PR #96, opened 2022-11-05
Enable/disable biometrics with fallback to pin

### LINKED ISSUE #75: Allow toggling of biometrics from Security screen

nop33:
Closes #75

Related to #30

This PR:
- Improves code readability
- Simplifies process of restoring navigation state even more (no more side-effects, only explicit actions trigger it)
- Fixes initialization bug in Addresses screen when switching wallets
- Matches secure storage configuration options with the one from popular mobile wallets
- Adds fallback to pin authentication when biometrics removed from device
- Adds biometrics toggle in Settings screen


nop33:
Just noticed that if I remove the fingerprints from my device there's an issue with fallback to pin, working on it...

nop33:
Fixed

src/screens/AddressesScreen.tsx
```diff
@@ -61,20 +62,58 @@ const AddressesScreen = ({ navigation }: ScreenProps) => {
...
+    () => navigation.navigate('ReceiveScreen', { addressHash: currentAddressHash }),
+    [currentAddressHash, navigation]
+  )
+
+  const navigateToEditAddressScreen = useCallback(
+    () => navigation.navigate('EditAddressScreen', { addressHash: currentAddressHash }),
+    [currentAddressHash, navigation]
+  )
+
+  const closeQRCodeModal = useCallback(() => setIsQrCodeModalOpen(false), [])
```
mvaivre, line 114: I fear that we may be going a bit too far with the use of `useCallback`, as this can be counter-productive in term of performances (and code complexity). We should make sure that this is needed.

Some interesting articles:
- https://kentcdodds.com/blog/usememo-and-usecallback
- https://dmitripavlutin.com/react-usecallback/

nop33: Thanks, I realized I need to revise this PR after what I learnt with the links you shared in the explorer PR!

nop33:
Ready for review again :) This PR simplifies a lot of code, to the point that adding the pin fallback results in fewer lines of code than before 😆  I feel clean and tidy!

src/storage/wallets.ts
```diff
@@ -150,51 +152,57 @@ export const deleteAllWallets = async () => {
...
   const walletsMetadata = await getWalletsMetadata()
   const existingWalletMetadata = walletsMetadata.find((wallet: WalletMetadata) => wallet.id === id)
 
-  if (existingWalletMetadata) {
-    Object.assign(existingWalletMetadata, partialMetadata)
+  if (!existingWalletMetadata) throw `Could not find wallet with ID ${id}`
 
-    await AsyncStorage.setItem('wallets-metadata', JSON.stringify(walletsMetadata))
-  }
+  Object.assign(existingWalletMetadata, partialMetadata)
```
mvaivre, line 195: As discussed, if we can avoid mutating object like that, would be great :)

PR #108, opened 2022-11-25
Improve navigation UX

nop33:
This PR improves 2 things:
- When switching wallets, if the user is already authenticated and the pin is in memory, there's no point in asking the user to enter it again. The pin is only asked when the app goes to the background and the memory is flushed.
- On a fresh install, while the user hasn't yet created a wallet, if the app went to the background the user had to restart the wallet creation process from scratch. I can imagine this would be a pain in the 🍑 if someone had to switch apps to see their mnemonic, see a pin stored in a password managed, etc. This PR solves that problem by updating and restoring the navigation state in these cases as well.


mvaivre (approved):
tiptop

PR #109, opened 2022-11-25
Import wallet by scanning QR code from desktop wallet

nop33:
The functionality for a proof of concept is here. ~~What remains is (as discussed on Slack) to figure out the issue with decrypting the encrypted mnemonic that is passed through the QR code from the desktop wallet to the mobile wallet.~~

Tested on Android.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,4 +20,4 @@ npm start
 
 Launch the Development Client app on your phone and enter the URL you see on your terminal (you can scan the QR code to get the URL).
 
-[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/2b937393-122d-4b37-b53a-4985e69a23d3
+[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/53835e97-e663-43fc-b6f3-8affc5a3c17f
--- a/README.md
+++ b/README.md
@@ -20,4 +20,4 @@ npm start
 
 Launch the Development Client app on your phone and enter the URL you see on your terminal (you can scan the QR code to get the URL).
 
-[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/53835e97-e663-43fc-b6f3-8affc5a3c17f
+[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/00080e0a-fca8-4b43-b988-e91e861227ed
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Install depedencies:
 npm i
 ```
 
-Install the [Development Client app][dev-client] (replacement for Expo Go).
+Install the [Development Client app](development-build.apk) on your Android device (replacement of the Expo Go mobile app).
 
 Start developer tools:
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -20,4 +20,4 @@ npm start
 
 Launch the Development Client app on your phone and enter the URL you see on your terminal (you can scan the QR code to get the URL).
 
-[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/2b937393-122d-4b37-b53a-4985e69a23d3
+[dev-client]: https://expo.dev/accounts/nop33/projects/mobile-wallet/builds/00080e0a-fca8-4b43-b988-e91e861227ed
```
mvaivre, line 23: Do we know how for long these dev builds are hosted?

patches/react-native-aes-crypto+2.1.1.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/node_modules/react-native-aes-crypto/android/src/main/java/com/tectiv3/aes/RCTAes.java b/node_modules/react-native-aes-crypto/android/src/main/java/com/tectiv3/aes/RCTAes.java
```
mvaivre, line 1: Super interesting to see how easy it is to patch a library.
This raises some questions though! Do you have any documentation you could share about patching?

- Will it be applied only if we use ver `2.1.1` of the library?
- What if we update the underlying library?
- How can we test this?

README.md
nop33: Good question. It appears that it's 30 days.

We could:
- Commit it to this repo
- Commit it to another repo

image

patches/react-native-aes-crypto+2.1.1.patch
nop33: The tool that was used was this: https://www.npmjs.com/package/patch-package.

Tha patches found in the `/patches` directory are automatically applied when installing the deps (`npm` and `yarn` pick them up).

In the _"Benefits of patching over forking"_ section, the docs mention:

> - Get told in big red letters when the dependency changed and you need to check that your fix is still valid.

So, to answer your first questions, it seems that it will always be applied and if the deps change it will show in the console so that we can check if it still works and/or update the patch.

How can we test this? :/ That's a good question...

README.md
mvaivre: Github is our go-to "hosting" service, so it would be great to use it for that.
Having them in this repo would make sense IMO, as they are required to actually be able to contribute to the project.

patches/react-native-aes-crypto+2.1.1.patch
mvaivre: Ok, that's good to know, thanks!

For now, as it's the only patch we have, I guess we can make sure we properly read the big red letters and manually test the impacted feature when updating the dep...

src/components/QRCodeScannerModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    getBarCodeScannerPermissions()
+  }, [])
+
+  const handleBarCodeScanned: BarCodeScannedCallback = ({ data }) => {
+    setScanned(true)
+    onQRCodeScan(data)
+    onClose()
+  }
+
+  if (!hasPermission) return null
```
mvaivre, line 53: What happens it he user declined to give the permissions by mistake? 
Will he/she be asked again thanks to getBarCodeScannerPermissions? Or will it just fail silently?

mvaivre (commented):
I can't test it right now cause I don't have my android phone with me but the code looks good! Just one question about permissions.
 
👏

src/components/QRCodeScannerModal.tsx
mvaivre: Yes, let's do that.

PR #113, opened 2022-12-06
Manual address discovery through Settings screen

nop33:


src/store/addressDiscoverySlice.ts
```diff
@@ -0,0 +1,207 @@
...
+
+          if (cachedData) {
+            if (cachedData.group !== group) {
+              continue
+            }
+
+            newAddressData = cachedData
+            newAddressGroup = cachedData.group
+          } else {
+            await sleep(1)
```
mvaivre, line 116: I've probably missed that in my previous review of this algo, but why is this sleep needed?

src/utils/addresses.ts
```diff
@@ -29,3 +30,83 @@ export const copyAddressToClipboard = (addressHash: AddressHash) => {
   Clipboard.setString(addressHash)
   Toast.show('Address copied!')
 }
+
+export const fetchAddressesData = async (addressHashes: AddressHash[]) => {
```
mvaivre, line 34: Discussion: I'm not sure about calling apis from a file in the `utils` folder. For me, those utilities only execute local code, without side effects. Question: where could this be stored instead?

src/store/addressesSlice.ts
```diff
@@ -172,84 +111,118 @@ export const fetchAddressesTransactionsNextPage = createAsyncThunk(
...
 
-    if (mainAddress && mainAddress.hash === newMainAddress.hash) {
-      throw 'Main address is already set to this address'
+      const addresses = addressesMetadata.map(({ index, ...settings }) => ({
+        ...deriveAddressAndKeys(masterKey, index),
+        settings
+      }))
+
+      dispatch(addressesAdded(addresses))
+      await dispatch(addressesDataFetched(addresses.map((address) => address.hash)))
```
mvaivre, line 135: I've started to reflect a bit on the past tense being used for dispatched actions. We start to use a lot of them, and I realise that it's sometimes a bit counterintuitive & too verbose for me.

Could we take some time soon for you to explain to me (again?) why you choose not to use the present form for this?

An initial thought that may make sense to me:
- use present for synchronous actions (like `flushPin` for instance)
- use past tense for async actions that may be awaited (like `await addressesDataFetched`). 

Not 100% sure either.

mvaivre (commented):
🤯 That's a biiig boy. 

First round, took me a while... I need a break :) Will come back to it soon ™️  after your answers.

src/store/addressDiscoverySlice.ts
nop33: This is done to allow the UI to react to user input while the address discovery is still taking place. This basically stops the execution of this function to allow other JS code to run (like the handling of a button press). I use it just before a blocking expensive operation. It is not necessary for discovering addresses, but I feel like it provides a nicer UX.

src/utils/addresses.ts
nop33: That's a good point, I also had doubts before moving it to utils and I am glad you are bringing it up. I'll try to think of a new proposal.

nop33: How about moving it inside the `src/api/client.ts` file, similarly to `buildSweepTransactions` and `signAndSendTransaction`?

nop33: This is what I propose:

<details><summary>See diff</summary>

```diff
diff --git a/src/api/client.ts b/src/api/client.ts
index 1028621..66dba47 100644
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -53,6 +53,45 @@ export class Client {
 
     return data
   }
+
+  async fetchAddressesData(addressHashes: AddressHash[]) {
+    const results = []
+
+    for (const addressHash of addressHashes) {
+      console.log('⬇️ Fetching address details: ', addressHash)
+      const { data } = await this.explorerClient.getAddressDetails(addressHash)
+      const availableBalance = data.balance
+        ? data.lockedBalance
+          ? (BigInt(data.balance) - BigInt(data.lockedBalance)).toString()
+          : data.balance
+        : undefined
+
+      console.log('⬇️ Fetching 1st page of address confirmed transactions: ', addressHash)
+      const { data: transactions } = await this.explorerClient.getAddressTransactions(addressHash, 1)
+
+      console.log('⬇️ Fetching address tokens: ', addressHash)
+      const { data: tokenIds } = await this.explorerClient.addresses.getAddressesAddressTokens(addressHash)
+
+      const tokens = await Promise.all(
+        tokenIds.map((id) =>
+          this.explorerClient.addresses.getAddressesAddressTokensTokenIdBalance(addressHash, id).then(({ data }) => ({
+            id,
+            balances: data
+          }))
+        )
+      )
+
+      results.push({
+        hash: addressHash,
+        details: data,
+        availableBalance: availableBalance,
+        transactions,
+        tokens
+      })
+    }
+
+    return results
+  }
 }
 
 const client = new Client(defaultNetworkSettings)
diff --git a/src/store/addressesSlice.ts b/src/store/addressesSlice.ts
index 7ce6495..8a6cfe7 100644
--- a/src/store/addressesSlice.ts
+++ b/src/store/addressesSlice.ts
@@ -32,7 +32,7 @@ import { getAddressesMetadataByWalletId } from '../storage/wallets'
 import { Address, AddressHash, AddressPartial, AddressSettings } from '../types/addresses'
 import { AddressToken } from '../types/tokens'
 import { PendingTransaction } from '../types/transactions'
-import { fetchAddressesData, storeAddressSettings } from '../utils/addresses'
+import { storeAddressSettings } from '../utils/addresses'
 import { getRandomLabelColor } from '../utils/colors'
 import { mnemonicToSeed } from '../utils/crypto'
 import { extractNewTransactions, extractRemainingPendingTransactions } from '../utils/transactions'
@@ -68,7 +68,7 @@ export const addressesDataFetched = createAsyncThunk(
     dispatch(loadingStarted())
 
     const addresses = payload
-    const results = await fetchAddressesData(addresses)
+    const results = await client.fetchAddressesData(addresses)
 
     dispatch(loadingFinished())
     return results
diff --git a/src/utils/addresses.ts b/src/utils/addresses.ts
index eaf7caf..c7507f7 100644
--- a/src/utils/addresses.ts
+++ b/src/utils/addresses.ts
@@ -19,7 +19,6 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 import * as Clipboard from 'expo-clipboard'
 import Toast from 'react-native-root-toast'
 
-import client from '../api/client'
 import { storeAddressMetadata } from '../storage/wallets'
 import { Address, AddressHash, AddressPartial, AddressSettings } from '../types/addresses'
 
@@ -31,45 +30,6 @@ export const copyAddressToClipboard = (addressHash: AddressHash) => {
   Toast.show('Address copied!')
 }
 
-export const fetchAddressesData = async (addressHashes: AddressHash[]) => {
-  const results = []
-
-  for (const addressHash of addressHashes) {
-    console.log('⬇️ Fetching address details: ', addressHash)
-    const { data } = await client.explorerClient.getAddressDetails(addressHash)
-    const availableBalance = data.balance
-      ? data.lockedBalance
-        ? (BigInt(data.balance) - BigInt(data.lockedBalance)).toString()
-        : data.balance
-      : undefined
-
-    console.log('⬇️ Fetching 1st page of address confirmed transactions: ', addressHash)
-    const { data: transactions } = await client.explorerClient.getAddressTransactions(addressHash, 1)
-
-    console.log('⬇️ Fetching address tokens: ', addressHash)
-    const { data: tokenIds } = await client.explorerClient.addresses.getAddressesAddressTokens(addressHash)
-
-    const tokens = await Promise.all(
-      tokenIds.map((id) =>
-        client.explorerClient.addresses.getAddressesAddressTokensTokenIdBalance(addressHash, id).then(({ data }) => ({
-          id,
-          balances: data
-        }))
-      )
-    )
-
-    results.push({
-      hash: addressHash,
-      details: data,
-      availableBalance: availableBalance,
-      transactions,
-      tokens
-    })
-  }
-
-  return results
-}
-
 export const findNextAvailableAddressIndex = (startIndex: number, skipIndexes: number[] = []) => {
   let nextAvailableAddressIndex = startIndex

 ```
</details>

File: src/store/addressDiscoverySlice.ts (hunk 1 above)
Comment (nop33) [reply src/store/addressDiscoverySlice.ts:116]: I had actually written a small comment on the other usage, but not on this one 😅 

https://github.com/alephium/mobile-wallet/blob/address-discovery/src/store/addressDiscoverySlice.ts#L84

File: src/store/addressesSlice.ts (hunk 3 above)
Comment (nop33) [reply src/store/addressesSlice.ts:135]: ## Redux recommended style guide

So, when I started learning about Redux and introducing it to the mobile wallet, I came across this recommendation and started using it directly:

> However, **we recommend trying to treat actions more as "describing events that occurred", rather than "setters".**
> https://redux.js.org/style-guide/#model-actions-as-events-not-setters

Their arguments are:
> Treating actions as "events" generally leads to:
> 1. more meaningful action names
> 2. fewer total actions being dispatched
> 3. a more meaningful action log history

As the style guide continues:

> **Actions should be written with meaningful, informative, descriptive type fields.** Ideally, you should be able to read through a list of dispatched action types, and have a good understanding of what happened in the application without even looking at the contents of each action.

## Personal opinion

Funnily enough, I have the complete opposite opinion of you, but I think I have good arguments to support it.

First of all, all of our current reducers/actions as well as our async thunks are written in past forms, as events that happened.

You proposed (tentatively) that reducers/actions should be written in the present tense and async thunks in past.

I propose that reducers/actions are written in past tense and the async thunks in present, and here's my reasoning:

Regarding argument (1), I cannot see how treating actions as events leads to more meaningful action names. The word "action" to me indicates that something is a "command", like "startLoading" or "fetchPrice". Also, when going through the code, it makes more sense to read `dispatch(startLoading())` rather than `dispatch(loadingStarted())` because my mental model around reducers/actions is (probably) wrong. I see them as functions, whereas Redux recommends to see them as events. So regarding this one, I actually don't care. I am fine to go with the recommended approach and tread them as events, using the past tense (especially due to point 3 though, not point 1).

Regarding argument (2), I also don't see any benefit 😅 I've read a couple times that we should avoid dispatching multiple actions one after the other and I try my best to reduce this and improve the code whenever I have a better idea. But there are a few times that I do dispatch 2 or 3 or sometimes 4 actions one after the other. I don't see how changing the tense of the reducer/action name changes anything.

Regarding argument (3) however, this is where I have a stronger opinion. Looking at the debugger, reading the list of "normal" actions that took place and their effect on the state makes a lot of sense (see the following example with `settings/generalSettingsChanged` followed by the actual change):

image

In the screenshot above I can see that "Oh, I see that the `price/statusChanged` to `updating` and then the `settings/generalSettingsChanged` from `'USD'` to `'CHF'`. It almost reads as natural English language.

What I have a slight problem with, however, is the async thunks (which is the reason I believe why you initially started this thread). See following example:

image

Reading `price/priceUpdated/pending` does not make much sense to me. It would make more sense and it would be more intuitive to read `price/updatePrice/pending`, then `price/statusChanged`, and finally `price/updatePrice/fullfiled`. Same for `addressDataFetched` vs `fetchAdddressData`:

image

The examples on async thunks in the RTK seem to support this idea, although they also use present tense for "normal" reducers/actions as well. https://redux-toolkit.js.org/api/createAsyncThunk

Sorry for the long comment, but I think it's worth having this discussion now! Looking forward to your thoughts!

File: src/screens/AddressDiscovery.tsx [hunk 4]
```diff
@@ -0,0 +1,196 @@
...
+const AddressDiscoveryScreen = ({ navigation }: ScreenProps) => {
+  const dispatch = useAppDispatch()
+  const theme = useTheme()
+  const addresses = useAppSelector(selectAllAddresses)
+  const discoveredAddresses = useAppSelector(selectAllDiscoveredAddresses)
+  const { loading, status } = useAppSelector((state) => state.addressDiscovery)
+  const [selectedAddressesToImport, setSelectedAddressesToImport] = useState<DiscoveredAddress[]>([])
+  const [importLoading, setImportLoading] = useState(false)
+
+  const startScan = () => dispatch(addressesDiscovered())
```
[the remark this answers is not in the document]
nop33: I believe our discussion below will resolve this issue (tl;dr: my proposal is to rename `addressesDiscovered` to `discoverAddresses`)

src/utils/addresses.ts
mvaivre: Putting this in `api` sounds good. 
And what about:
- moving `buildSweepTransactions ` and `signAndSendTransaction ` out of `client.ts` to `src/api/transactions.ts` (they don't really need to be attached to the client class IMO)
- moving `fetchAddressesData`to `src/api/addresses`

WDYT?

src/store/addressesSlice.ts
nop33: Continuing my research on the topic, I think I really like this opinion. It's worth the read: https://medium.com/magnetis-backstage/why-action-is-a-bad-name-for-a-redux-action-68bec375539e

Specifically, this hit a chord in me:

> Actions are a payload of information that sends data from your application to your store. [...] It's the one and only form of communication between the application and your Store. [...] What should my application be telling my store? [...] Most projects show a mix of the two: communication of intent and communication of an event. [...] the application should inform the Store of what's happening and the Store will know what to do [...] actions are a mean of communication with the outside world. [...] It’s not action, it’s a reaction. It’s not giving orders, it’s describing behavior. It’s not imperative, it’s declarative. And that’s why Action is a bad name for a Redux Action.

I think I have already introduced many anti-patterns in our code-base, probably because I am thinking of actions as functions...

nop33: My only concern with treating actions as functions and not as events is that we are being led into redux antipatterns... Everywhere in the fundamentals of redux I read:

> The actions, the events that occur in the app based on user input, and trigger updates in the state

> You can think of an action as an event that describes something that happened in the application.

> You can think of dispatching actions as "triggering an event" in the application. Something happened, and we want the store to know about it. Reducers act like event listeners, and when they hear an action they are interested in, they update the state in response.

https://redux.js.org/tutorials/fundamentals/part-2-concepts-data-flow

I would like to continue the research on "why treating actions as functions and not as events leads to antipatterns".

nop33: What I mentioned just above, btw, is also in line with your comment above: https://github.com/alephium/mobile-wallet/pull/113#discussion_r1046016355

> It's a bit weird to me that `startScan` is dispatching `addressesDiscovered` instead of `addressDiscoveryStarted` 🤔
What's the rational behind this?

I think you're absolutely right. `startScan` should dispatch `scanStarted` (or `addressDiscoveryStarted`, whatever) and let the `addressDiscoverySlice` react to this event by starting the address discovery algo.

_Continues research..._

src/utils/addresses.ts
nop33: Sounds like a great idea man, thanks!

nop33: I took care of this in the following PR because I was working on a lot of refactoring, I hope that's ok!

nop33:
I've opened a new PR with the refactoring I made to apply the new Redux mindset that I acquired. Is it ok if we merge this one? 😇

mvaivre:
yes :)

PR #116, opened 2022-12-16
Redux mindset

### LINKED ISSUE #111: Custom network settings form fields are not initialized

nop33:
I'm sorry for the big PR @mvaivre, but it was almost impossible to break this into chunks. I was constantly refactoring as I was learning and many changes depended on others. Nevertheless, we're still under the artificial PR limit of 1000 LOC 😅 

This PR:
1. applies a major refactoring on how we update our Redux state. We are moving from calling actions in an imperative form and treating them as "commands" to using a declarative form of treating them as "events that happened in our app that the state needs to know about".
2. reduces the use of async middleware and moved the code into simple async functions that the components can call
3. moves API related functions into the `src/api/` dir
4. fixes the some implicit state duplication (anti-pattern) in preference of (memoized) selectors (the way-to-go solution to avoid implicit state duplication)
5. fixes #111 
6. renames a ton of things

There are a couple of things that imo could be improved. Specifically:
- I don't like that we depend on `useEffect` to do the client and address initialization in `src/hooks/useInitializeClient.ts`. I've added comments with questions on how we could do better.
- I've commended some anti-patterns that need to be fixed soon, but I won't do this in this PR cause it'll grow even bigger

Resources that helped me navigate:
- [Redux Anti-Patterns - Part 1. State Management.](https://blog.mgechev.com/2017/12/07/redux-anti-patterns-race-conditions-state-management-duplication/)
- [Why Action is a bad name for a Redux Action](https://medium.com/magnetis-backstage/why-action-is-a-bad-name-for-a-redux-action-68bec375539e)
- [The Perils of Using a Common Redux Anti-Pattern](https://itnext.io/the-perils-of-using-a-common-redux-anti-pattern-344d778e59da)
- [Style Guide | Redux](https://redux.js.org/style-guide)
- [Actions | Redux](https://redux.js.org/faq/actions#is-there-always-a-one-to-one-mapping-between-reducers-and-actions)


src/hooks/useRefreshALPHPrice.ts
```diff
@@ -18,20 +18,20 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { useEffect } from 'react'
 
-import { priceUpdated } from '../store/priceSlice'
+import { updatePrice } from '../store/priceSlice'
 import { useAppDispatch, useAppSelector } from './redux'
 
 const useRefreshALPHPrice = () => {
```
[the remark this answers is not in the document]
mvaivre: Related: wouldn't it make more sense to dispatch some actions when initializing the App, instead of using hooks?

In App.tsx currently:

```ts
const Main = ({ children }: { children: ReactNode }) => {
  useInitializeClient()
  useLoadStoredSettings()
  useAppStateChange()
  useRefreshALPHPrice()

  return <>{children}</>
}
```

We could have one `useEffect` to dispatch one action `initializeApp` calling simple functions to initialize the client and load the stored settings. 

For the `useRefreshALPHPrice()` I think that the polling mechanism should be implemented "closer" to where the query is made (see comment above).

I'm less sure about `useAppStateChange()`.

src/hooks/useSwitchWallet.ts
```diff
@@ -0,0 +1,40 @@
...
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { changeActiveWallet, deriveAddressesFromStoredMetadata } from '../persistent-storage/wallets'
+import { walletUnlocked } from '../store/activeWalletSlice'
+import { ActiveWalletState } from '../types/wallet'
+import { useAppDispatch } from './redux'
+
+const useSwitchWallet = () => {
```
mvaivre, line 24: It's a bit weird to me to use a hook for this. I'm not 100% sure why. 

Maybe because the only reason it's a hook is because we use `useAppDispatch`.
However, it's possible to import `dispatch` our of a component by importing the store.. So this could technically be a util function instead of a hook.

src/persistent-storage/wallets.ts
```diff
@@ -16,24 +16,55 @@ You should have received a copy of the GNU Lesser General Public License
...
+    firstAddress: {
+      index: 0,
+      hash: wallet.address,
+      publicKey: wallet.publicKey,
+      privateKey: wallet.privateKey
+    }
+  }
+}
+
+const persistWallet = async (
```
mvaivre, line 67: Linked to comment above: `switchWallet` could be declared in this file instead of being a hook.

src/screens/LoginScreen.tsx
```diff
@@ -48,20 +47,12 @@ const LoginScreen = ({
...
-      if (resetWalletOnLogin || addressesStatus === 'uninitialized') {
-        dispatch(addressesFromStoredMetadataInitialized())
-      }
+      const requiresAddressInitialization = resetWalletOnLogin || addressesStatus === 'uninitialized'
+      await switchWallet(wallet, requiresAddressInitialization, pin)
 
       restoreNavigationState(resetWalletOnLogin)
     },
-    [addressesStatus, dispatch, resetWalletOnLogin, restoreNavigationState]
+    [addressesStatus, resetWalletOnLogin, restoreNavigationState, switchWallet]
```
mvaivre, line 55: This goes in the direction of my previous argument about not getting `switchWallet` from a hook : IMO it makes no sense that `switchWallet` is part of the deps here, as the function won't change (it doesn't rely on state).

src/screens/SettingsScreen.tsx
```diff
@@ -48,26 +48,40 @@ const SettingsScreen = ({ navigation }: ScreenProps) => {
...
   ] = useAppSelector((s) => [s.settings, s.network.name, s.activeWallet])
-  const biometricsEnabled = authType === 'biometrics'
 
-  const toggleBiometrics = async () => await dispatch(biometricsToggled({ enable: !biometricsEnabled }))
+  const isBiometricsEnabled = activeWallet.authType === 'biometrics'
+
+  const toggleBiometrics = async () => {
+    if (isBiometricsEnabled) {
+      await disableBiometrics(activeWallet.metadataId)
+      dispatch(biometricsDisabled())
```
mvaivre, line 62: Yeah! This makes sense, and really simplify the state logic.

src/screens/SwitchWalletScreen.tsx
```diff
@@ -61,23 +62,22 @@ const SwitchWalletScreen = ({ navigation, style }: SwitchWalletScreenProps) => {
...
+      let mnemonic = storedWallet.mnemonic
 
       if (storedWallet.authType === 'pin') {
         if (pin) {
           const decryptedWallet = await walletOpenAsyncUnsafe(pin, storedWallet.mnemonic, pbkdf2, mnemonicToSeed)
-          await dispatch(activeWalletChanged({ ...storedWallet, mnemonic: decryptedWallet.mnemonic }))
-          dispatch(addressesFlushed())
-          dispatch(addressesFromStoredMetadataInitialized())
-          restoreNavigationState(true)
+          mnemonic = decryptedWallet.mnemonic
```
mvaivre, line 70: OMG this diff <3 so much simpler 💯

src/screens/new-wallet/PinCodeCreationScreen.tsx
```diff
@@ -96,7 +98,9 @@ const PinCodeCreationScreen = ({ navigation }: ScreenProps) => {
     if (method === 'create') {
       setLoading(true)
 
-      await dispatch(walletGeneratedAndStoredWithPin({ name, pin: pinCode }))
+      const wallet = await generateAndStoreWallet(name, pinCode)
+      dispatch(newWalletGenerated(wallet))
+      dispatch(syncAddressesData([wallet.firstAddress.hash]))
```
mvaivre, line 103: Again: so much better 👏

mvaivre (commented):
👏 Awesome evolution, great job! As said on Slack, here are some comments that may hopefully trigger some interesting discussions.

src/hooks/useInitializeClient.ts
```diff
@@ -19,33 +19,50 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
       console.error('Could not connect to network: ', network.name)
       console.error(e)
-      dispatch(networkStatusChanged('offline'))
     }
-  }, [dispatch, network.name, network.settings])
+  }, [network.settings.nodeHost, network.settings.explorerApiHost, network.name, dispatch])
 
+  // Is there a better way to trigger the initial client initialization?
+  // Currently we trigger it "magically" by setting the the networkSlice status status to 'connecting', which is
+  // happening when loading the stored network settings and when network settings are updated by the user.
```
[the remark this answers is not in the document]
nop33: > I'd say that initializing the client should be the first thing done once the app is up and renaming - wether the network is up or not.
If the network isn't working for whatever reason, an error should be dispatched accordingly from the initialization function.

So far, so good, we got this already.

> Thinking out loud : in order to decouple the client initialization from state changes, maybe we shouldn't use a hook for this. We should be more explicit and dispatch an initialization action from the places that should trigger the initialization process, instead of listening for state changes from the hook...

This is exactly what I wanted to do and I spent a few hours on it, without finding a proper solution. I was working on the form that edits the network settings. When the user clicks "Save custom network settings", I want to trigger a new network initialization and if it's successful, I want to fetch the data for each address. This is easy, I can just run this code right after the network settings are saved, wait for it to succeed, and then fetch the address data:

```ts
const initializeClient = async () => {
  try {
    await client.init(network.settings.nodeHost, network.settings.explorerApiHost)
    dispatch(apiClientInitSucceeded())
  } catch (e) {
    dispatch(apiClientInitFailed())
  }
}
```

Here's the problem though: One of our requirements is that the app SHOULD keep trying to initialize the client in case the client initialization has failed. I have not found a way to do that AND also remember that the app should fetch the address once it succeeds. I tried using callbacks, but I was not satisfied with my solution because it doesn't work for every case. The only way that works as it should is using 2 `useEffect`s that listen to state changes: one that keeps trying to initialize the client if the network is offline, and one that initializes the address data if their status is "uninitialized" and the network is "online".

> App has started => Run the client initialisation.
If the client isn't available, call the init function again from the network status change listener.

That's what we do at the moment with:

```tsx
useEffect(() => {
  if (network.status === 'connecting') {
    initializeClient()
  }
}, [initializeClient, network.status])

const shouldInitialize = network.status === 'offline'
useInterval(initializeClient, 2000, !shouldInitialize)
```

> And for clarity sake, it would maybe be beneficial to decouple the network state from the client state.
So creating a new "Client" slice may help?

Can't see on top of my head how this could help...

src/hooks/useRefreshALPHPrice.ts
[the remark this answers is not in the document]
nop33: > Would it make sense to add the refresh mechanism as a redux middleware?

Hm, from my understanding, a middleware is like a "man in the middle" that intercepts a fired action to do some (some times async) processing of the action, before it goes to the reducer. I am not sure how much sense it makes to dispatch an action that will dispatch other actions in an inteval...

> Related, here's what the RTK Query docs say about polling: https://redux-toolkit.js.org/rtk-query/usage/polling
> ...
> For the useRefreshALPHPrice() I think that the polling mechanism should be implemented "closer" to where the query is made (see comment above).



This is interesting, I remember implementing a PoC on the desktop wallet to poll the price from coingecko.

If we use the RTK Query, then we won't need to store the price in the Redux store. We currently need the price in 3 components:
- `src/screens/ReceiveScreen.tsx`
- `src/components/TransactionRow.tsx`
- `src/components/BalanceSummary.tsx`

I wonder, which of these components should have the polling mechanism? One of them or all of them? If it's one of them, how easy will it be to find which one it is when we look for it? If all of them, it doesn't sound very DRY...

> Related: wouldn't it make more sense to dispatch some actions when initializing the App, instead of using hooks?
> We could have one useEffect to dispatch one action initializeApp calling simple functions to initialize the client and load the stored settings.

Just for clarity, having 1 useEffect is not better than have many, necessarily. 1 useEffect would be that the dependency array would grow, as the code inside of it grows. This means that we'd need to add if/else statements to check what changed in order to control what code should run. 

That being said, I am trying to think how we could restructure the app initialization to do what you suggest. Here's my initial thoughts:
- Regarding `useInitializeClient`, we have a discussion above where I explain that I did not manage to avoid the 3 `useEffects` (one for the initial client init, one for the interval, and one for fetching address data). Let's continue the discussion there to find a better solution, if any.
- Regarding `useLoadStoredSettings`, it was initially extracted in a hook so that the `Main` component looks tidy. It includes a `useEffect` that runs only once, which is the desired behavior. We could move it into an async middleware, but then we would need to `dispatch` its action from somewhere. This needs to be inside a `useEffect`. Which is basically how it is now.
- Regarding `useAppStateChange`, I'm also less sure, like you. There's one `useEffect` listening to app state changes. Mixing it with the above would be confusing imo.
- Regarding `useRefreshALPHPrice`, we are discussing it just above. I'd like to find a better solution as well, if there's any/

src/hooks/useSwitchWallet.ts
nop33: Haha, I also didn't like it 100% when I implemented it. So I tried a different approach: To not use a hook, and just be mor explicit with dedicated dispatched actions, based on exactly what is happening inside the app. This is the code I wrote. I pushed it to a diff branch so that it's not confused with this one: https://github.com/alephium/mobile-wallet/commit/2db898e0f89e641b9fe8d92948eb06e52484a130

src/persistent-storage/wallets.ts
nop33: I think my comment above answers this comment. https://github.com/alephium/mobile-wallet/pull/116#discussion_r1053085646

src/screens/LoginScreen.tsx
nop33: Yes, it's there for the same reason `dispatch` was there.

src/screens/SettingsScreen.tsx
nop33: Nice! Glad you see it the same way I do!

src/hooks/useSwitchWallet.ts
nop33: What do you think of this solution? https://github.com/alephium/mobile-wallet/commit/44ae680fe9934a1cb7474ff194e3aaf0b8051013

src/persistent-storage/wallets.ts
nop33: Probably fixed in https://github.com/alephium/mobile-wallet/commit/44ae680fe9934a1cb7474ff194e3aaf0b8051013 😄

src/screens/LoginScreen.tsx
nop33: Probably fixed by https://github.com/alephium/mobile-wallet/commit/44ae680fe9934a1cb7474ff194e3aaf0b8051013

src/hooks/useSwitchWallet.ts
nop33: FWIT, discussion continued here: https://github.com/alephium/mobile-wallet/commit/44ae680fe9934a1cb7474ff194e3aaf0b8051013#r93654507

mvaivre:
Looking good! One - hopefully last - [suggestion](https://github.com/alephium/mobile-wallet/commit/05545a6e9608c614ea2969994cc9f92215b99ed3#r93675876) :)

nop33:
@mvaivre Following out conversation today, let me know if I can merge this :)

mvaivre:
LFG!

mvaivre (approved):

src/hooks/useInitializeClient.ts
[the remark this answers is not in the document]
mvaivre: NOTE: Following our discussion, we will in the future investigate the potential use of the RTK Query that may allow us to better manager retries and polling.

src/hooks/useRefreshALPHPrice.ts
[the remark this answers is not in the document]
mvaivre: NOTE: Following our discussion, we will in the future investigate the potential use of the RTK Query that may allow us to better manager retries and polling.

PR #129, opened 2023-05-16
Fix splash screen

### LINKED ISSUE #106: Stuck on the landing screen
Clue : putting the app in the background then restoring it allows to pass this step.

nop33:
Closes #106

In this PR I did manly 3 things:
- Got rid of the `useAppStateChange` hook and instead moved its code directly in the `App.tsx` file. I find it less confusing this way to understand what are the "first things" that happen in the app. Similar approach has been followed on the desktop wallet lately.
- Got rid of the `useRestoreNavigationState` hook so that the depedencies of other hooks is clearer. It was not very clear to me that some `useEffects` were being re-executed because the deps of the `useRestoreNavigationState` hook were updated.
- Added a "special" hook that runs only 1 time and executes the same code as the one being executed when the app becomes active. For some reason, the `'change'` event of the `AppState` is not triggered on the first app launch. I chose to disable eslint for the deps of that hook, otherwise it gets executed every time the `lastNavigationState` changes. I did not find an easy alternative solution. I added a TODO to revisit this approach.

### DIFF
--- a/src/screens/LoginScreen.tsx
+++ b/src/screens/LoginScreen.tsx
@@ -27,7 +27,7 @@ import { deriveWalletStoredAddresses, rememberActiveWallet } from '../persistent
 import { walletSwitched, walletUnlocked } from '../store/activeWalletSlice'
 import { AddressPartial } from '../types/addresses'
 import { ActiveWalletState } from '../types/wallet'
-import { setNavigationState } from '../utils/navigation'
+import { resetNavigationState, setNavigationState } from '../utils/navigation'
 
 type ScreenProps = StackScreenProps<RootStackParamList, 'LoginScreen'>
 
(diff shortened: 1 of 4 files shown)


src/screens/LoginScreen.tsx
```diff
@@ -59,19 +58,19 @@ const LoginScreen = ({
         addressesToInitialize = await deriveWalletStoredAddresses(wallet)
 
         dispatch(walletSwitched({ wallet, addressesToInitialize, pin }))
-        restoreNavigationState(true)
+        setNavigationState()
```
mvaivre, line 61: Nit: what about setting it to `null` explicitly? Or better yet (IMO), make `setNavigationState` accept either a new state as argument (as it is now), OR the string 'initial' to make it clear that we want to reset the state. Or better yet (IMO) 🙃 to use a `resetNavigationState` function?

mvaivre (commented):
One tiny comment. 👏

src/screens/LoginScreen.tsx
nop33: hehe alright, I'll make it more explicit! I agree with you

mvaivre (approved):

PR #131, opened 2023-05-17
Use web3

### LINKED ISSUE #126: Use web3 and latest sdk

nop33:
I'm so glad this works so smoothly!

Closes #126


mvaivre (approved):
Apart from the "not so optimal web3 export structure", this was indeed surprisingly painless! LGTM.

PR #137, opened 2023-05-17
Fix tx loading and reduce re-renders

nop33:


src/utils/addresses.ts
```diff
@@ -95,9 +95,14 @@ export const getAddressAvailableBalance = (address: Address): bigint =>
 export const selectAddressTransactions = (
```
mvaivre, line 95: NIT: Not part of the PR I know, but this follows the selector naming scheme, so I'd expect this to be a standalone redux selector. I know it's used in multiple slices, but I find it weird that it's not located in the "store" folder. WDYT?

nop33: Good point. I think our file structure needs to be adapted to match the new one in the desktop wallet (which is still not set in stone actually). I'll add a comment about this in the next PR.

src/screens/TransfersScreen.tsx
```diff
@@ -17,32 +17,27 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 type ScreenProps = StackScreenProps<InWalletTabsParamList & RootStackParamList, 'TransfersScreen'>
 
 const TransfersScreen = ({ navigation }: ScreenProps) => {
-  const addressHashes = useAppSelector(selectAddressIds) as AddressHash[]
-  const [confirmedTransactions, pendingTransactions, haveAllPagesLoaded] = useAppSelector((s) => [
-    selectAddressesConfirmedTransactions(s, addressHashes),
-    selectAddressesPendingTransactions(s, addressHashes),
-    selectHaveAllPagesLoaded(s)
-  ])
+  const selectAddressesConfirmedTransactions = useMemo(makeSelectAddressesConfirmedTransactions, [])
```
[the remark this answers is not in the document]
mvaivre: Well I still don't really like to mix factories with "pure" selectors, but it's really not a big deal :) 
I have to admit that there's one "pros" to this approach: creating factories for heavy selectors forces us to always use them through useMemo.

mvaivre (approved):

PR #138, opened 2023-05-23
Use token list to fetch token metadata

nop33:


mvaivre (approved):
👏  

It's true that a lot is shared with the desktop wallet. If the utils / redux functions are considered "stable", a shared internal library could start to make sense.

nop33:
> If the utils / redux functions are considered "stable"

Such a philosophical question 

💀🤔💭

PR #139, opened 2023-05-23
Use path prefix

nop33:


babel.config.js
```diff
@@ -20,6 +20,17 @@ module.exports = function (api) {
...
     presets: ['babel-preset-expo'],
-    plugins: ['react-native-reanimated/plugin']
+    plugins: [
+      'react-native-reanimated/plugin',
+      [
+        'module-resolver',
+        {
+          root: ['./src'],
+          alias: {
+            '~': './src'
```
nop33, line 30: You might be asking yourself (and rightfully so) "why the hell did he use `~` instead of `@`" and I wouldn't blame you. It would simply not be able to resolve paths when using `@`. Actually, it would, but with a VERY WEIRD issue: if there is an empty line between the last import and the code of the file, it would not be able to import the first import... Go figure. I thought there was something wrong with my config, but once I tried other characters, it worked.

I looked up several resources without being able to pin-point the issue. So, `~`. 😅 

- https://github.com/tleunen/babel-plugin-module-resolver
- https://medium.com/mtholla/add-alias-paths-and-module-resolution-to-your-react-native-project-7caab7ef4799
- https://koprowski.it/import-alias-in-react-native-and-vscode/
- https://github.com/facebook/react-native/issues/27522
- https://stackoverflow.com/questions/59699226/react-native-with-typescript-unable-to-resolve-modules
- https://www.reactnativeschool.com/how-to-setup-path-alias-in-a-react-native-typescript-app

mvaivre: WTF 🙃 I guess `~` it is then! Thanks for sharing the resources.

mvaivre (approved):

PR #140, opened 2023-05-24
Handle display of swap txs and cleanup UI

nop33:


nop33:
@mvaivre just a casual ping 🏓

ios/Podfile.properties.json
```diff
@@ -1,4 +1,4 @@
 {
-  "expo.jsEngine": "hermes",
+  "expo.jsEngine": "jsc",
```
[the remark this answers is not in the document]
mvaivre: That's super weird and very unfortunate. Hermes brings substantial performance improvements. I was able to debug using Hermes before, so I'll take a look.
Have you tried using flipper?

src/App.tsx
```diff
@@ -67,29 +67,27 @@ dayjs.updateLocale('en', {
...
+          <RootStackNavigation />
+          <StatusBar style={theme.name === 'light' ? 'dark' : 'light'} />
+        </ThemeProvider>
+      </Main>
+    </Provider>
   )
 }
 
-const Main = ({ children }: { children: ReactNode }) => {
+const Main: FC = ({ children }) => {
```
mvaivre, line 90: Nit: I seem to remember that we decided not to rely on `FC` anymore, like [many others](https://backstage.io/docs/architecture-decisions/adrs-adr006/#:~:text=Facebook%20has%20removed%20React.,combination%20with%20a%20few%20downsides.).

src/components/Badge.tsx
```diff
@@ -17,26 +17,22 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 }
 
-const Badge = ({ style, color, children }: BadgeProps) => {
-  return (
-    <View style={style}>
-      <BadgeText color={color}>{children}</BadgeText>
-    </View>
-  )
-}
+const Badge: FC<BadgeProps> = ({ style, color, children }) => (
```
mvaivre, line 31: Why are we back to using FC :(

src/components/ExpandableRow.tsx
```diff
@@ -28,12 +28,11 @@ import HighlightRow from '~/components/HighlightRow'
 interface ExpandableRowProps {
   // TODO: Find a better way to measure the height of the collapsable section
   expandedHeight: number
-  children: ReactNode
   title?: string
   style?: StyleProp<ViewStyle>
 }
 
-const ExpandableRow = ({ expandedHeight, children, title = 'Advanced options', style }: ExpandableRowProps) => {
+const ExpandableRow: FC<ExpandableRowProps> = ({ expandedHeight, children, title = 'Advanced options', style }) => {
```
mvaivre, line 35: Same as before. I'll stop commenting those from now on.

src/style/style.d.ts
```diff
@@ -51,6 +52,7 @@ declare module 'styled-components' {
       valid: string
       star: string
       pale: string
+      complementary
```
mvaivre, line 55: No type?

src/types/components.d.ts
```diff
@@ -16,7 +16,4 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { Asset } from '@alephium/sdk'
-
-export const sortAssetsByName = (tokenA: Asset, tokenB: Asset) =>
-  !tokenA.name || !tokenB.name ? tokenA.id.localeCompare(tokenB.id) : tokenA.name.localeCompare(tokenB.name)
+type FC<P = unknown> = React.FC<React.PropsWithChildren<P>>
```
mvaivre, line 19: 🤔

src/utils/transactions.ts
```diff
@@ -16,27 +16,90 @@ You should have received a copy of the GNU Lesser General Public License
...
+
+  let amount: bigint | undefined = BigInt(0)
+  let direction: TransactionDirection
+  let infoType: TransactionInfoType
+  let outputs: explorer.Output[] = []
+  let lockTime: Date | undefined
+  let tokens: Required<AssetAmount>[] = []
+
+  if (isPendingTx(tx)) {
+    direction = 'out'
```
mvaivre, line 54: Note: we will need to make this more flexible (cf. explorer) as pending TXs could be incoming as well.

mvaivre (approved):
I approve, but let's talk about this FC thingy :)

ios/Podfile.properties.json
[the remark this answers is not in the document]
nop33: I remember that I have tried in the past, without success. If flipper has a nice redux integration, I could give it another try.

src/App.tsx
nop33: For clarification, this comes from here: https://github.com/alephium/mobile-wallet/blob/swap/src/types/components.d.ts#L19

Last thing I remember was that we use `FC` when we need the `children` prop to be there, and not when we don't. But that was for the desktop wallet, I don't remember if we had discussed this on the mw side.

I have not encountered any disadvantages using `FC`. If anything, it makes our code-bases more consistent. We don't need to switch the mental model in our brain that says "when I want to use `children` I just need to use `FC`.

I look forward to your thoughts!

src/style/style.d.ts
nop33: lol, how the hell did i miss this 😆

src/components/Badge.tsx
nop33: I thought it was a mistake that we forgot to update so I changed everything :D

src/App.tsx
nop33: https://github.com/facebook/create-react-app/pull/8177

ios/Podfile.properties.json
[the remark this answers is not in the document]
mvaivre: They have plugins! https://javascript.plainenglish.io/how-to-debug-redux-in-react-native-using-flipper-d785b46cf7bf

src/App.tsx
mvaivre: I personally like that we type the children prop manually. This makes it very explicit that we're actually expecting children. It also allows us to be more strict when we need to, ie. accept only children of a certain type, accept only 1 child, not 2 etc.

And it's funny, because for me using FC makes our component declaration less consistent. Sometimes we use it, sometimes we don't. I like when we always have the full list of typed props defined just above the component itself. 😊

mvaivre: If this is fine by you, then yeah! 🚀

ios/Podfile.properties.json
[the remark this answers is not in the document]
nop33: btw, I think this only affects the simulators and not the actual production build. I could be wrong though. So far I get very good performance with the simulator.

nop33:
I know you approved, but I re-requested a review after adding 2 more commits :)

PR #141, opened 2023-05-25
WIP: Start using react-native-gifted-charts

nop33:
I like this library. I like that it's interactive and that it was relatively easy to create a chart that resembles the one in Figma. I found some bugs, however. I've created a [PR](https://github.com/Abhinandan-Kushwaha/react-native-gifted-charts/pull/321) to fix them, let's see if it gets addressed. For now, I simply patched the library.

@mvaivre if you have a strong opinion against this lib and for another one, lemme know!

image


nop33:
What I don't like though is that our type checker finds many type errors in the package...

src/components/HistoricWorthChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+          date,
+          worth: price * parseFloat(toHumanReadableAmount(totalAmountPerDate))
+        }
+      })
+    }
+
+    const trimInitialZeroDataPoints = (data: DataPoint[]) => data.slice(data.findIndex((point) => point.worth !== 0))
+
+    let dataPoints = computeChartDataPoints()
+    dataPoints = trimInitialZeroDataPoints(dataPoints)
```
mvaivre, line 105: Nit: 

```ts
const dataPoints = trimInitialZeroDataPoints(computeChartDataPoints())
```

Also, `computeChartDataPoints` is only used once and doesn't need to be a function?

mvaivre (approved):
Just a nit! 👍

src/components/HistoricWorthChart.tsx
nop33: Both `computeChartDataPoints` and `trimInitialZeroDataPoints` are called only once. The reason I made them functions is to use the function name as a hint to the developer of what is going on there so that I don't have to add a comment. I think it improves code readability.

mvaivre: Yes, I agree that declaring `trimInitialZeroDataPoints ` makes it more readable. IMO `chartDataPoints` could be a variable, but I'm 💯%  fine if we keep a function for this.

nop33: How about dis? :) https://github.com/alephium/mobile-wallet/pull/141/commits/294c42d8771774cd9e67cfbf09a4b658e0c2f597

mvaivre: Yes! Just be careful, you have a styled component hidden in between to utils functions!

nop33: oopsi!

PR #143, opened 2023-06-01
Addresses tabs

nop33:
Regarding the tabs, there might be some insights once we figure out the differences between:
- react-native-pager-view
- react-native-tab-view
- @react-navigation/material-top-tabs


src/components/AppText.tsx
```diff
@@ -22,13 +22,16 @@ export interface AppTextProps {
   bold?: boolean
   semiBold?: boolean
   medium?: boolean
-  color?: keyof DefaultTheme['font'] | keyof DefaultTheme['global']
+  color?: keyof DefaultTheme['font'] | keyof DefaultTheme['global'] | string
```
mvaivre, line 25: Should we allow this to be a generic string? Wouldn't this fail bellow (line 35), if I provide a random string instead of a key from the theme object?

We could have 2 fields: `color` which accepts a generic string, and `themeColor`?

And if we want to be strict with regard to typings, we could use discriminating union like: 

```ts
type ColorProps = {
  color: keyof DefaultTheme['font'] | keyof DefaultTheme['global'];
  themeColor?: never;
};

type ThemeColorProps = {
  color?: never;
  themeColor: keyof DefaultTheme['font'] | keyof DefaultTheme['global'];
};

type NoColorProps = {
  color?: never;
  themeColor?: never;
};

type ComponentProps = ColorProps | ThemeColorProps | NoColorProps;
```

src/components/Carousel.tsx
```diff
@@ -31,6 +33,8 @@ interface CarouselProps<T> {
   distance?: number
   onScrollStart?: () => void
   onScrollEnd?: (index: number) => void
+  FooterComponent?: ReactNode
+  style?: StyleProp<ViewStyle>
 }
 
 function Carousel<T>({
```
mvaivre, line 40: Note: A matter of personal pref, but if we want to be consistent and use arrow function, we can do 

```ts
const Carousel = <T,>() => ....
```

mvaivre (commented):
🏅

src/components/AppText.tsx
nop33: maybe instead of `themeColor` we could have `theme`

nop33: Sth like this:

```ts
import { themes, ThemeType } from '~/style/themes'

type FontColor = keyof DefaultTheme['font']
type GlobalColor = keyof DefaultTheme['global']

export type AppTextProps = {
  bold?: boolean
  semiBold?: boolean
  medium?: boolean
  color?: FontColor | GlobalColor
  colorTheme?: ThemeType
  size?: number
}

export default styled.Text<AppTextProps>`
  color: ${({ color, theme, colorTheme }) => {
    const th = colorTheme ? themes[colorTheme] : theme

    return color ? th.font[color as FontColor] || th.global[color as GlobalColor] : th.font.primary
  }};
`
```

mvaivre: Yes, this would work. In terms of naming, maybe `colorTheme` could be replaced by `forcedTheme` or sth like this.

nop33: Hm, I like `colorTheme`, it directly connects to the `color` prop since they have the same prefix. Do you have a strong opinion about it?

src/components/Carousel.tsx
nop33: Fixed!

src/components/AppText.tsx
mvaivre: No strong opinion :)

mvaivre (approved):

mvaivre:
🚀 great job!

PR #144, opened 2023-06-07
Basic send functionality

nop33:
I just realized how big this PR started to get so I'll stop here and continue with the following things on a follow-up PR:
- Add Send/Receive button in Dashboard screen
- Skip Origin screen when navigating through the send button of the Addresses screen


src/components/headers/BottomModalHeader.tsx
nop33, line 1: Most of the file's code is copied from `src/hooks/layout/useBottomModalOptions.tsx` with the addition of `children` so that the send navigation can append its own header to the basic bottom modal header.

src/contexts/SendContext.tsx
```diff
@@ -0,0 +1,210 @@
...
+        Toast.show(getHumanReadableError(e, 'Could not send transaction'))
+      }
+    },
+    [address, assetAmounts, consolidationRequired, dispatch, toAddress, unsignedTxData.unsignedTxs]
+  )
+
+  const authenticateAndSend = useCallback(
+    async (onSendSuccess: () => void) => {
+      if (requiresAuth) {
+        setOnSendSuccessCallback(() => onSendSuccess)
```
nop33, line 172: This little `() =>` was the solution to the problem I mentioned on Slack today ;)

src/contexts/SendContext.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const [unsignedTxData, setUnsignedTxData] = useState<UnsignedTxData>({ unsignedTxs: [], fees: initialValues.fees })
+
+  const [consolidationRequired, setConsolidationRequired] = useState(false)
+  const [isConsolidateModalVisible, setIsConsolidateModalVisible] = useState(false)
+  const [isAuthenticationModalVisible, setIsAuthenticationModalVisible] = useState(false)
+  const [onSendSuccessCallback, setOnSendSuccessCallback] = useState<() => void>(() => () => null)
+
+  const address = useAppSelector((s) => selectAddressByHash(s, fromAddress ?? ''))
+
+  const setAssetAmount = (assetId: string, amount?: bigint) => {
```
mvaivre, line 87: Idea: what about using a `Map<AssetId, bigint>` for this? This would make the get / set processes that much easier! It would also remove the need for storing `set` functions in the global context.

About using Maps with React: 
- https://usehooks-ts.com/react-hook/use-map
- https://azimi.io/es6-map-with-react-usestate-9175cd7b409b

src/screens/Send/AssetsScreen.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+export default AssetsScreen
+
+interface AssetRowProps {
+  asset: Asset
+  isLast: boolean
+  style?: StyleProp<ViewStyle>
+}
+
+const AssetRow = ({ asset, style, isLast }: AssetRowProps) => {
```
mvaivre, line 95: What about doing the same as in the DW, ie. having a `AssetsScreen` folder, with an index file and specific components (one file per component)?

src/screens/Send/AssetsScreen.tsx
```diff
@@ -0,0 +1,269 @@
...
+      amountValueAsFloat > parseFloat(availableAmount)
+        ? 'Amount exceeds available balance'
+        : asset.id === ALPH.id && amountValueAsFloat < parseFloat(minAmountInAlph) && amountValueAsFloat !== 0
+        ? `Amount must be greater than ${minAmountInAlph}`
+        : tooManyDecimals
+        ? `This asset cannot have more than ${asset.decimals} decimals`
+        : ''
+
+    setError(newError)
+    if (newError) return
```
mvaivre, line 128: Very nit: I'd add a blank before this line 🙃

src/contexts/SendContext.tsx
mvaivre: I know what you're going to say, that we're used to specific `set` functions when using `setState`, so it makes sense to build the context in a similar way. But I find this idea interesting anyway: we never use Maps, however we often have cases where we manipulate arrays of key/values...

src/screens/Send/SendScreenHeader.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <ContinueButton onPress={() => navigation.navigate('OriginScreen')} />
+        )}
+      </SendScreenHeaderStyled>
+    </BottomModalHeader>
+  )
+}
+
+export default SendScreenHeader
+
+export const BackButton = (props: PressableProps) => {
```
mvaivre, line 69: This could probably be extracted and reused in other places.

src/screens/Send/SendScreenHeader.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <ChevronLeft size={25} color={theme.global.accent} />
+    </BackButtonStyled>
+  )
+}
+
+interface ContinueButtonProps extends PressableProps {
+  text?: string
+}
+
+export const ContinueButton = ({ text = 'Continue', ...props }: ContinueButtonProps) => (
```
mvaivre, line 83: Same for this. We could group those components in a `components/buttons` folder.

src/contexts/SendContext.tsx
nop33: Just to be clear, you recommend to instead of using `assetAmounts: AssetAmount[]` array in the `SendContextProps` to use a `Map<AssetId, bigint>`?

nop33: > It would also remove the need for storing set functions in the global context.

I don't understand this part!

---

Thanks for the links!

src/screens/Send/SendScreenHeader.tsx
nop33: Also, are we ready to start DRYing the buttons code yet, or are the designs going to evolve you think? For example, the "Continue" button could possibly be the same component as the "Scan" and "Paste" ones, or the "New address" 

image

image

image

image

but probably a different component than

image

image

or maybe they could all be unified, but our `Buttons` component needs a lot of cleanup 😅 

I was kind waiting for more stable designs before proceeding with this, to avoid too early optimization.

nop33: 🏓  @mvaivre

mvaivre: I'd wait and dry things up when cleaning up the UI before release!

PR #162, opened 2023-06-23
Contacts CRUD

nop33:
Sorry for the long PR, the feature of contacts is a big one. I will work on the remaining sub-features (like ~contact sharing,~ selecting contact in send screens #160, etc) in a follow-up PR.


src/persistent-storage/wallets.ts
```diff
@@ -260,6 +262,70 @@ export const persistAddressesMetadata = async (walletId: string, addressesMetada
   await persistWalletsMetadata(walletsMetadata)
 }
 
+export const persistContact = async (contactData: ContactFormData) => {
```
mvaivre, line 265: Nit: why not adding those function to a dedicated `contacts.ts` file?

mvaivre:
Functional testing: I get an error when trying to create a contact.

src/persistent-storage/wallets.ts
nop33: Good question. I chose to leave them in this file because I made `contacts` be part of the wallet metadata object. Since all wallet metadata functions are in this file, I decided to keep these here as well. Not a strong opinion on this.

nop33:
> Functional testing: I get an error when trying to create a contact.

I fixed this, it was probably cause the wallet you used had old metadata stored. I updated the code to take this into consideration.

src/screens/Addresses/Contact/EditContactScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+          onPress={() => {
+            Alert.alert('Deleting contact', 'Are you sure you want to delete this contact?', [
+              { text: 'Cancel' },
+              {
+                text: 'Delete',
+                onPress: async () => {
+                  setIsDeleting(true)
+
+                  try {
+                    await deleteContact(params.contactId)
```
mvaivre, line 63: This function call should always be followed by the dispatch of `contactDeletedFromPersistentStorage` right? In this case, why not dispatch directly from within the `deleteContact` function?

src/screens/Addresses/Contact/EditContactScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+    })
+  }, [dispatch, navigation, params.contactId])
+
+  if (!contact) return null
+
+  const handleSavePress = async (formData: ContactFormData) => {
+    setLoading(true)
+
+    try {
+      await persistContact(formData)
```
mvaivre, line 87: Same remark as above.

src/screens/Addresses/Contact/NewContactScreen.tsx
```diff
@@ -0,0 +1,69 @@
...
+    address: ''
+  }
+
+  const handleSavePress = async (formData: ContactFormData) => {
+    setLoading(true)
+
+    try {
+      const id = await persistContact(formData)
+
+      dispatch(contactStoredInPersistentStorage({ ...formData, id }))
```
mvaivre, line 51: Here too (I won't signal the others if any)

src/components/ListItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+interface ListItemProps extends PressableProps {
+  title: string
+  subtitle: string | ReactNode
+  icon: ReactNode
+  isLast?: boolean
+  style?: StyleProp<ViewStyle>
+  innerStyle?: StyleProp<ViewStyle>
+  hideSeparator?: boolean
+  rightSideContent?: ReactNode
+  children?: ReactNode
```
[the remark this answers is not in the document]
nop33: It does, but if I remove it I get this error:

```
    Type 'ReactNode | ((state: PressableStateCallbackType) => ReactNode)' is not assignable to type 'ReactNode'.ts(2769)
```

nop33:
I've addressed all your comments in a follow-up PR, is that a problem for you?https://github.com/alephium/mobile-wallet/pull/167/commits

mvaivre (approved):
🏄‍♂️ 👏

PR #164, opened 2023-07-18
Display tokens and NFTs in horizontal lists

nop33:


mvaivre:
I sooo want to try this with a big wallet so I tries with the infamous "devnet" wallet 🙃 Unfortunately this seems to get the wallet stuck into infinite loading.

nop33:
Maybe you can try it with a smaller wallet (import the devnet wallet and skip the address discovery) and I can make a new PR for loading optimization.

nop33:
I didn't get in an infinite loading btw, but it took a while

image

nop33:
I've created a new PR to improve the loading experience: #165

src/components/UnknownTokensListItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
```
mvaivre, line 18: image

A note for our future us on this component's design: let's not use a ListItem as it may be confusing when quickly looking at the list IMO. 

I'll redesign it late. Could you add a little todo in the code to make sure I don't miss it? Thanks!

src/screens/SendReceive/AddressBox.tsx
```diff
@@ -33,8 +33,8 @@ interface AddressBoxProps extends PressableProps {
 }
 
 const AddressBox = ({ addressHash, isSelected, ...props }: AddressBoxProps) => {
-  const selectAddressesAssets = useMemo(makeSelectAddressesAssets, [])
-  const assets = useAppSelector((s) => selectAddressesAssets(s, [addressHash]))
+  const selectAddressesKnownFungibleTokens = useMemo(makeSelectAddressesKnownFungibleTokens, [])
```
mvaivre, line 36: Nit: what about separating the useAppSelector calls from the selector factories with a new line?

nop33: I kinda like keeping them close to each other, they are tightly coupled. If you have a strong opinion about this I can change it, but I'd like to keep it as it is.

mvaivre: Okay, very nit anyway :)

src/components/UnknownTokensListItem.tsx
nop33: TODO added!

src/store/assets/assetsActions.ts
```diff
@@ -18,28 +18,85 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+export const loadingStarted = createAction('assets/loadingStarted')
+
+export const syncNetworkTokensInfo = createAsyncThunk(
+  'assets/syncNetworkTokensInfo',
+  async (_, { getState, dispatch }) => {
+    const state = getState() as RootState
+
+    dispatch(loadingStarted())
+
+    let metadata = undefined
```
mvaivre, line 39: Nit: this could just be

```ts
let metadata
```

mvaivre (commented):
All good!

mvaivre (approved):

PR #167, opened 2023-08-04
WIP: Tabs

nop33:


mvaivre:
Very very nice! Starting with functional testing:

- The header style is updating even though the screen is not focused: 

https://github.com/alephium/mobile-wallet/assets/6715943/8c462aaf-666a-45d1-8e93-13cf92b33bbf

The header style change also glitch a bit when a modal opens (see vid).

- Let's use a vertical grid for NFTs instead of a horizontal scrollable section as we now have plenty of space 🌠

nop33:
Regarding your video, this was an issue also from before and not related to this PR. I created an issue to track it and tackle it in a follow up PR if that's ok: https://github.com/alephium/mobile-wallet/issues/168

> Let's use a vertical grid for NFTs instead of a horizontal scrollable section as we now have plenty of space 🌠

I'll work on that in this PR! 2 NFTs per row should be enough, right? I'll try to figure out how other apps are doing it.

mvaivre:
The grid looks good!

mvaivre (approved):
Nothing to say on this one, LGTM!

PR #181, opened 2023-08-15
Configure app name, app icons, builds

### LINKED ISSUE #178: Choose app name
Currently, our app is named `mobile-wallet` (defined in `app.json`: https://github.com/alephium/mobile-wallet/blob/master/app.json#L3).

Shouldn't we call it something else? How about simply "Alephium"? "Alephium wallet" might be a bit too long and "Alephium mobile wallet" even longer. Other ideas?

### LINKED ISSUE #179: Create an app icon

nop33:
1. Updates icons (Closes #179)
2. Updates app name (and defines a different name for development client by converting `app.json` to `app.config.js` and using env vars) (Closes #178)
3. Prepares `eas submit` configuration (Depends on #186)


app.config.js
```diff
@@ -0,0 +1,68 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+const IS_DEV = process.env.APP_VARIANT === 'development'
+
+export default {
+  expo: {
+    name: IS_DEV ? 'Alephium Wallet (DEV)' : 'Alephium Wallet',
```
mvaivre, line 23: Interesting. Is this a recommended approach? (same remark regarding the different bundleIdentifiers)

I've never seen this before, usually it's the same app name, but different release pipelines (for alpha, beta etc.).

nop33: I saw this in the expo docs. 

https://docs.expo.dev/build-reference/variants/

It doesn't affect alpha, beta, etc. It's for us to be able to have 2 apps on our devices, both prod (for personal use) and dev (for, well, development).

mvaivre (approved):

PR #188, opened 2023-08-21
Modalize

nop33:
The ~3.5K additions are auto-generated code that doesn't need a review 😉


src/components/layout/Modalize.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { ComponentProps, forwardRef } from 'react'
+import { Modalize as RNModalize } from 'react-native-modalize'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+
+const Modalize = forwardRef(function Modalize({ children, ...props }: ComponentProps<typeof RNModalize>, ref) {
```
mvaivre, line 23: NIT: we could use an arrow function notation.

src/components/layout/ScrollFlatListScreen.tsx
```diff
@@ -28,17 +27,6 @@ interface ScreenProps<T> extends FlatListProps<T> {
 
 const ScrollFlatListScreen = <T,>({ style, onScrollEndDrag, ...props }: ScreenProps<T>) => {
   const listRef = useRef<FlatList<T>>(null)
-  const navigation = useNavigation()
-
-  useScrollToTop(listRef) // Scrolls to top when tapping the active tab
```
mvaivre, line 33: This makes me think that we could implement that in the refactored DefaultHeader => a click in an empty area in the header could scroll to the top, like in the X app.

src/components/layout/TabScrollScreen.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
```
mvaivre, line 18: The only difference between this component and `ScrollScreen` is that it scrolls to the top when blurred? 
If yes, should we maybe keep only 1 configurable `ScrollScreen`?

src/components/layout/ScrollFlatListScreen.tsx
nop33: I think that's a great idea!

src/screens/Address/NewAddressScreen.tsx
```diff
@@ -80,14 +87,19 @@ const NewAddressScreen = ({ navigation }: ScreenProps) => {
   }
 
   return (
-    <>
-      <AddressForm
-        initialValues={initialValues}
-        onSubmit={handleGeneratePress}
-        onGroupPress={() => navigation.navigate('GroupSelectScreen')}
-      />
+    <Modal {...props}>
```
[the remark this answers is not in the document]
nop33: Absolutely correct, a mistake on my side!

src/screens/Address/GroupSelectModal.tsx
```diff
@@ -17,34 +17,35 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import { ScreenSection } from '~/components/layout/Screen'
-import ScrollScreen from '~/components/layout/ScrollScreen'
+import ScrollModal from '~/components/layout/ScrollModal'
+import { ScrollScreenProps } from '~/components/layout/ScrollScreen'
 import RadioButtonRow from '~/components/RadioButtonRow'
 import { useNewAddressContext } from '~/contexts/NewAddressContext'
-import { NewAddressNavigationParamList } from '~/navigation/NewAddressNavigation'
 
-type ScreenProps = StackScreenProps<NewAddressNavigationParamList, 'GroupSelectScreen'>
+interface GroupSelectModalProps extends ScrollScreenProps {
```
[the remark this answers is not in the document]
nop33: I grouped our modals into one file (`Modals.tsx`) and created a generic type. I think you'll like this approach!

```ts
export type ModalProps<ScreenType = ScreenProps | ScrollScreenProps> = ScreenType & {
  onClose?: () => void
}
```

src/components/layout/Modalize.tsx
nop33: When we converting it to arrow function we get this eslint error:


> Component definition is missing display name (eslint [react/display-name](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/display-name.md))

To avoid it, we either disable the rule or need to manually set the display name with:

```ts
Modalize.displayName = 'Modalize'
```

The complete code would look like this:

```ts
const Modalize = forwardRef<RNModalize, ComponentProps<typeof RNModalize>>(({ children, ...props }, ref) => {
   ...
})
Modalize.displayName = 'Modalize'
```

I personally prefer the named function:

```ts
const Modalize = forwardRef(function Modalize({ children, ...props }: ComponentProps<typeof RNModalize>, ref) {
  ...
})
```

mvaivre: Alright!

src/components/layout/ScrollFlatListScreen.tsx
mvaivre: Awesome. (Merging this branch into my `UI-refinement` one will be a pain in the 🍑 😆)

src/components/layout/TabScrollScreen.tsx
mvaivre: Mhh, we could just use the code from `TabScrollScreen` (renaming the component `ScrollScreen`), and simply check the condition (`scrollToTopOnBlur`) in the useEffect? 

One other issue I have with this `TabScrollScreen` is : "how is its implementation related to Tabs"? :)

src/screens/Addresses/AddressesScreen.tsx
```diff
@@ -111,13 +131,10 @@ const AddressesScreen = ({ navigation, style, route: { params } }: ScreenProps)
             scrollTo={scrollToCarouselPage}
             FooterComponent={
               <>
+                {/* TODO: Do we need this button if we only have 1 or 2 addresses?   */}
```
mvaivre, line 134: I would say no. Let's show it from 3 addresses on.

src/screens/SendReceive/Send/DestinationScreen.tsx
```diff
@@ -88,6 +94,32 @@ const DestinationScreen = ({ navigation, style, route: { params } }: ScreenProps
...
+    }
+  }
+
+  const handleAddressPress = (toAddressHash: AddressHash) => {
+    posthog?.capture('Send: Selected own address to send funds to')
+
+    navigation.navigate('SendNavigation', {
+      screen: nextScreen,
+      params: { toAddressHash }
+    })
```
mvaivre, line 118: For addresses and contacts: let's not directly navigate to the next screen on select.

As a user, I find it important to be able to check directly if the right address has been selected (it's easy to do an accidental tap).

Idea: would be to create a little animation in the "destination address" field, when it's automatically filled up. This would attract the user's attention and encourage them to quickly check the address hash.

Do you think you could create such an animation? Don't take too much time for this as I may tweak it : for now, let's just make the input's background smoothly fade to the accent color and back to the original color (1s total). WDYT? :)

src/screens/SettingsScreen.tsx
```diff
@@ -114,59 +121,69 @@ const SettingsScreen = ({ navigation }: ScreenProps) => {
...
+          <ButtonStyled
+            title="Add a new wallet"
+            Icon={PlusIcon}
+            variant="valid"
+            onPress={() => navigation.navigate('LandingScreen')}
+          />
+          <ButtonStyled title="Delete this wallet" Icon={Trash2} variant="alert" onPress={handleDeleteButtonPress} />
+        </ScreenSection>
+      </ScrollScreen>
+      <Portal>
```
mvaivre, line 178: Nit: what about adding a line break before the portal section?

src/screens/Address/NewAddressScreen.tsx
```diff
@@ -80,14 +87,21 @@ const NewAddressScreen = ({ navigation }: ScreenProps) => {
...
 
   return (
-    <>
-      <AddressForm
-        initialValues={initialValues}
-        onSubmit={handleGeneratePress}
-        onGroupPress={() => navigation.navigate('GroupSelectScreen')}
-      />
+    <Screen {...props}>
+      <NewAddressContextProvider>
```
mvaivre, line 91: Do we really need to use a context here? 🤔 

Couldn't the group be passed directly from the GroupSelectModal to the parent?

nop33: I am removing it in my follow up work on WC, I hope you don't mind if I don't touch this one.

mvaivre: Also, I've just seen during my testing that the group doesn't seem to update properly.

src/screens/SendReceive/Send/DestinationScreen.tsx
nop33: I've pushed a commit! I used the `global.pale` color cause I didn't like the animation with the `bg.accent` one. It looks weird, I think because of the opacity level. Feel free to tweak it.

mvaivre: Yeah I expected that! Good choice.

mvaivre (approved):
Great work 👏 👏

PR #195, opened 2023-09-01
Integrate WalletConnect

nop33:


src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -0,0 +1,715 @@
...
+
+  const onSessionExtend = useCallback((args: SignClientTypes.EventArguments['session_extend']) => {
+    console.log('📣 RECEIVED EVENT TO EXTEND SESSION')
+    console.log('👉 ARGS:', args)
+  }, [])
+
+  const onProposalExpire = useCallback((args: SignClientTypes.EventArguments['proposal_expire']) => {
+    console.log('📣 RECEIVED EVENT TO EXPIRE PROPOSAL')
+    console.log('👉 ARGS:', args)
+  }, [])
```
nop33, line 397: I've added these one simply to help me figure out what the hell I am receiving from WC, these are all the events I could find in the docs.

src/components/buttons/ButtonStack.tsx
```diff
@@ -25,11 +25,7 @@ interface ButtonStackProps extends ViewProps {
 }
 
 const ButtonStack = ({ children, style }: ButtonStackProps) => (
-  <View style={style}>
-    {children?.map((child, i) => (
-      <ButtonContainer key={i}>{child}</ButtonContainer>
-    ))}
-  </View>
+  <View style={style}>{children?.map((child, i) => <ButtonContainer key={i}>{child}</ButtonContainer>)}</View>
```
nop33, line 28: Linting was broken for some reason so after upgrading the eslint related packages, this changes was required (also for next file)

src/components/QRCodeScannerModal.tsx
```diff
@@ -174,6 +176,11 @@ const ScreenSectionCentered = styled(ScreenSection)`
   align-items: center;
 `
 
+const TextContainer = styled.View`
+  width: 80%;
+  border: 0px solid transparent; // This is a hack cause I don't freaking understand why the text doesn't expand
```
nop33, line 181: 😭

src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -0,0 +1,714 @@
+/*
```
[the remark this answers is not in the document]
mvaivre: This file seems to be a rather "simple" in terms of code (mostly conditions and state update)... I feel like all we need now is a lot of testing.

src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -0,0 +1,714 @@
...
+      if (!walletConnectClient) return
+
+      console.log('⏳ RESPONDING TO WC WITH:', { topic: event.topic, response })
+      await walletConnectClient.respond({ topic: event.topic, response })
+      console.log('✅ RESPONDING: DONE!')
+    },
+    [walletConnectClient]
+  )
+
+  const respondToWalletConnectWithSuccess = async (event: SessionRequestEvent, result: SignResult) => {
```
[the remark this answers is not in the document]
nop33: Because it's only used by `handleApprovePress` which is not used in a dependency array of anything, so there's no need to wrap with a `useCallback`.

mvaivre:
This is some heavy stuff sir

mvaivre:
My propositions: 

 - [ ] Switch the address that is used: open a bottom modal on top (feasible?) to select an address on the fly
- [ ] Highlight the connection button using accent color

nop33:
>  Switch the address that is used: open a bottom modal on top (feasible?) to select an address on the fly

It's quite tricky to implement that, I would need to lift the state of the proposal modal up to the context and hack it through, it didn't feel clean at all. Since the modal works well when its content is changed, I'd rather leave it as it is for now (I improved the layout a bit)

mvaivre (approved):

PR #202, opened 2023-09-12
Fix contact list

nop33:
@mvaivre I am making a PR on top of yours so that it's easy for you to review (instead of pushing directly on your branch) and so that we don't stumble on each other's feet.


mvaivre:
Functional testing: I still don't see any animation on close when using the address or contact selector in the Destination screen.
Also, I've seen that `handleClose` is now called twice which may bring unexpected side effects.

src/screens/Addresses/AddressesScreen.tsx
```diff
@@ -59,8 +59,8 @@ const AddressesScreen = ({ onScroll, contentStyle, ...props }: BottomBarScrollSc
   const [selectedAddressHash, setSelectedAddressHash] = useState(defaultAddress?.hash ?? '')
   const selectedAddress = useAppSelector((s) => selectAddressByHash(s, selectedAddressHash))
 
-  const [quickSelectionModalOpen, setQuickSelectronModalOpen] = useState(false)
-  const [addressSettingsModalOpen, setAddressSettingsModalOpen] = useState(false)
+  const [isQuickSelectionModalOpen, setIsQuickSelectronModalOpen] = useState(false)
```
mvaivre, line 62: That's on me but: quickSele**ctro**n => quickSelection :P

nop33:
> Functional testing: I still don't see any animation on close when using the address or contact selector in the Destination screen.

Hm, I do:


https://github.com/alephium/mobile-wallet/assets/1579899/7671a20f-13e4-4565-9053-b636740ff21e

> Also, I've seen that handleClose is now called twice which may bring unexpected side effects.

I am not very happy with the solution either. I would like to discuss some things.

nop33:
@mvaivre ok, I properly fixed the Contacts modal closing animation when navigating to a new screen, without calling `handleClose` multiple times. Here's what it looks like, super smooth:


https://github.com/alephium/mobile-wallet/assets/1579899/d8b13c68-8eb6-49fa-87bd-711d4b8aed16

mvaivre (approved):
This looks great 🙏 Awesome job, thanks.

PR #205, opened 2023-09-13
Scrollable modals

nop33:
- Every `ModalContent` is now an instance of `ScrollView` (from gesture handler)
- A new component called `ModalFlatListContent` can be used instead of `ModalContent` for rendering modals that include a `FlatList` (also from gesture handler)
- Unless the `maximisedContent` or the `customMinHeight` prop is set, the `BottomModal` will take the height of its contents.

~~This doesn't work well for minimized modals. Needs investigation.~~ It does!

### DIFF
--- a/src/components/layout/ModalContent.tsx
+++ b/src/components/layout/ModalContent.tsx
@@ -17,33 +17,35 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { ReactNode } from 'react'
-import { ViewProps } from 'react-native'
+import { FlatListProps, ScrollViewProps } from 'react-native'
+import { FlatList as GHFlatList, ScrollView as GHScrollView } from 'react-native-gesture-handler'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
-import styled from 'styled-components/native'
 
 import Screen, { ScreenProps } from '~/components/layout/Screen'
 import ScrollSection, { ScrollSectionProps } from '~/components/layout/ScrollSection'
 import { VERTICAL_GAP } from '~/style/globalStyle'
 
-export interface ModalContentProps extends ViewProps {
-  onLayout: ViewProps['onLayout']
+interface ModalContentBaseProps {
   onClose?: () => void
   isScrollable?: boolean
   children?: ReactNode
   verticalGap?: number | boolean
-  fill?: boolean
 }
 
-export const ModalContent = ({ children, verticalGap, fill, ...props }: ModalContentProps) => (
-  <ModalContentStyled
-    {...props}
-    style={{
-      gap: verticalGap ? (typeof verticalGap === 'number' ? verticalGap || 0 : VERTICAL_GAP) : 0,
-      flex: fill ? 1 : undefined
-    }}
-  >
+export interface ModalContentProps extends ModalContentBaseProps, ScrollViewProps {}
+
+export interface ModalFlatListContentProps<T> extends ModalContentBaseProps, FlatListProps<T> {}
+
+export const ModalContent = ({ children, verticalGap, ...props }: ModalContentProps) => (
+  <GHScrollView {...getDefaultProps({ verticalGap })} {...props}>
     {children}
-  </ModalContentStyled>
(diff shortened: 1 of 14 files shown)


src/components/layout/ModalContent.tsx
```diff
@@ -66,7 +68,15 @@ export const ScrollModal = ({ children, style, ...props }: ScrollSectionProps) =
   )
 }
 
-const ModalContentStyled = styled.View`
-  padding-top: 10px;
-  padding-bottom: 10px;
-`
+const getDefaultProps = ({ verticalGap, contentContainerStyle }: ModalContentProps) => ({
```
mvaivre, line 71: Nit: the `contentContainerStyle` properties isn't used in this file (only `verticalGap? is passed to `getDefaultProps`). You wanted to keep it for potential future use?

nop33: It is being used in line 80!

nop33: ah sorry, my bad

nop33: Updated! 😅

mvaivre:
A couple of things I'm noticing:
- I still don't have any closing animation when selecting an address or a contact in the Destination screen (not a huge deal). After closing, I don't see any opening animation either.
- I notice a jump down when swiping the Transaction modal up to expand. As if it first quickly wants to minimise before expanding. When swiping down to minimise, it's the contrary. It seems to first expand then minimize.
- The transaction modal lacks the vertical gap (margin below title)
- There's a right margin next to the address selection modal opened from the AddressesScreen. 

Other than this it works well. One last suggestion: we could maybe stop using `maximisedContent` for the currency and network settings in the Settings modal. WDYT?

nop33:
> I still don't have any closing animation when selecting an address or a contact in the Destination screen (not a huge deal). After closing, I don't see any opening animation either.

Fixed ✅ 

> The transaction modal lacks the vertical gap (margin below title)

Fixed ✅ 

> There's a right margin next to the address selection modal opened from the AddressesScreen.

Fixed ✅ 

> One last suggestion: we could maybe stop using maximisedContent for the currency and network settings in the Settings modal.

Done ✅ 

> I notice a jump down when swiping the Transaction modal up to expand. As if it first quickly wants to minimise before expanding. When swiping down to minimise, it's the contrary. It seems to first expand then minimize.

This bug already exists in the base branch (you can test it by setting a `customMinHeight={600}` to the Transaction modal. I think I know what the reason is:

This is the code of handling the layout change of the modal:

```ts
Gesture.Pan()
        .onStart((e) => {
          offsetY.value = modalHeight.value
        })
        .onChange((e) => {
          if (position.value !== 'closing') {
            modalHeight.value = offsetY.value + e.translationY
          }
        })
        .onEnd(() => {
          const shouldMinimise = position.value === 'maximised' && -modalHeight.value < dimensions.height - DRAG_BUFFER

          const shouldMaximise =
            canMaximize.value && position.value === 'minimised' && -modalHeight.value > minHeight.value + DRAG_BUFFER

          const shouldClose =
            ['minimised', 'closing'].includes(position.value) && -modalHeight.value < minHeight.value - DRAG_BUFFER

          if (shouldMaximise) {
            handleMaximize()
          } else if (shouldMinimise) {
            scrollableContent ? handleClose() : handleMinimize()
          } else if (shouldClose) {
            handleClose()
          } else {
            modalHeight.value =
              position.value === 'maximised'
                ? withSpring(-maxHeight, springConfig)
                : withSpring(-minHeight.value, springConfig)
          }
        }),
```

When the user lifts their finger the `onEnd` gets triggered. However, until the code of `onEnd` runs to decide what should happen (maximization, minimization, etc), for a split of a second the modal tries to bring itself back to its previous position, but then one of the `handleMaximize`, `handleMinimize` gets called. The solution I think is that the decision for maximizing/minimizing should be taken in the `onChange` callback and not in the `onEnd`. The `onEnd` should only apply the decision. WDYT?

mvaivre (approved):
That's perfect. Let's merge this!

PR #207, opened 2023-09-14
Use scaled down left and right components in compact header

mvaivre:
I tried multiple implementations, but that's the easiest, lightest & most practical I came with.
This doesn't allow for customisation, but it works for now IMO. WDYT?


nop33 (approved):
Works well

PR #209, opened 2023-09-15
Fix jump when expanding/minimizing

mvaivre:


mvaivre:
@nop33 In the end it was just an issue of non-animated padding change. Going ahead and merging in order to clean things up.

PR #210, opened 2023-09-17
Header tap scroll top

mvaivre:


src/contexts/ScrollContext.tsx
```diff
@@ -16,26 +16,31 @@ You should have received a copy of the GNU Lesser General Public License
...
 
+type ScrollableView = ScrollView | FlatList
+
+export type ScrollDirection = 'up' | 'down' | null
+export type ScrollableViewRef = RefObject<ScrollView> | RefObject<FlatList>
+
 interface ScrollContextValue {
   scrollY?: SharedValue<number>
   scrollDirection?: SharedValue<ScrollDirection>
+  activeScreenRef?: MutableRefObject<ScrollableView | null>
```
nop33, line 31: Smart idea!!!

src/contexts/ScrollContext.tsx
```diff
@@ -16,26 +16,31 @@ You should have received a copy of the GNU Lesser General Public License
...
 */
 
-import { createContext, ReactNode, useContext } from 'react'
+import { createContext, MutableRefObject, ReactNode, RefObject, useContext, useRef } from 'react'
+import { FlatList, ScrollView } from 'react-native'
 import { SharedValue, useSharedValue } from 'react-native-reanimated'
 
+type ScrollableView = ScrollView | FlatList
+
+export type ScrollDirection = 'up' | 'down' | null
```
nop33, line 25: Should we get rid of this since we are not using it and clean up this component, or leave it for potential future use?

src/hooks/layout/useScrollToTopOnFocus.ts
```diff
@@ -17,17 +17,17 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import { useNavigation } from '@react-navigation/native'
-import { RefObject, useEffect } from 'react'
-import { FlatList, ScrollView } from 'react-native'
+import { useEffect } from 'react'
 
+import { ScrollableViewRef } from '~/contexts/ScrollContext'
 import { scrollScreenTo } from '~/utils/layout'
 
-const useScrollToTopOnFocus = (viewRef: RefObject<ScrollView> | RefObject<FlatList>) => {
+const useScrollToTopOnFocus = (viewRef: ScrollableViewRef) => {
```
nop33, line 25: I wanted to mention this before as well but I forgot: I find the name of this function a bit confusing when I look at its implementation. From what I read in the docs, `blur` is the event fired when focus is **removed** from a screen. Why did you go for this naming, what am I missing?

nop33 (commented):
Looks good! Smart idea with the active screen ref!

I only found out this issue, not sure it's related to this PR tho

https://github.com/alephium/mobile-wallet/assets/1579899/3a126c62-2f79-4ed4-afda-dfe5fb63847f

src/contexts/ScrollContext.tsx
mvaivre: We may need to use it soon and it's not too much noise. Let's keep it for a few iterations? If we see then that we don't need it, we'll remove it.

src/hooks/layout/useScrollToTopOnFocus.ts
mvaivre: That's absolutely correct :P I wanted to scroll to top on focus initially, but ended up changing the event to blur while forgetting to update the fn name 😆

mvaivre:
> I only found out this issue, not sure it's related to this PR tho

Shit. Seems like it's also present in the base UI-refinements branch.

nop33 (approved):

PR #216, opened 2023-09-18
UI refinements - Standalone headers for most of screens (solves nav header glitches)

nop33:


src/components/headers/NavigationStackHeader.tsx
```diff
@@ -0,0 +1,36 @@
+/*
```
mvaivre, line 1: So we have 3 really similar Headers: NavigationBaseHeader, NavigationStackHeader, and StackHeader!
TBH I find this a bit confusing. Maybe we could find a way to make it dryer?

src/hooks/layout/useScreenNavigationScrollHandler.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import { useFocusEffect } from '@react-navigation/native'
+import { useCallback } from 'react'
+import { NativeScrollEvent, NativeSyntheticEvent } from 'react-native'
+
+import { ScrollableViewRef, useNavigationScrollContext } from '~/contexts/NavigationScrollContext'
+
+const scrollDirectionDeltaThreshold = 10
+
+const useScreenNavigationScrollHandler = (viewRefForScrollTopOnHeaderPress?: ScrollableViewRef) => {
```
mvaivre, line 27: Nit: Long name :) What about `useNavigationScrollHandler`, which works well with `useNavigationScrollContext`?

src/hooks/layout/useScreenScrollHandler.tsx
```diff
@@ -16,50 +16,20 @@ You should have received a copy of the GNU Lesser General Public License
...
 
-import { scrollEndThreshold } from '~/components/headers/BaseHeader'
-import { ScrollableViewRef, useScrollContext } from '~/contexts/ScrollContext'
+const useScreenScrollHandler = () => {
+  const [screenHeaderHeight, setScreenHeaderHeight] = useState(0)
+  const screenScrollY = useSharedValue(0)
 
-const scrollDirectionDeltaThreshold = 10
+  const screenScrollHandler = (e: NativeSyntheticEvent<NativeScrollEvent>) =>
+    (screenScrollY.value = e.nativeEvent.contentOffset.y)
```
mvaivre, line 28: Nit: Isn't that weird to use parenthesis instead of brackets?

mvaivre (commented):
I like it!!! It feels much more natural to set the header options in the Screen props 👍 
Also, performances seems fine!

One tiny thing I have spotted, we lack a top margin in the pin screen :)

src/components/headers/NavigationStackHeader.tsx
nop33: As discussed on Slack, since the use-cases for each of those headers is quite different, I didn't find a way to make it DRYer than it already is. `NavigationBaseHeader` is using the `BaseHeader` and only has the minimal necessary code to make it work with the scroll context and navigation props. The `NavigationStackHeader` naturally needs to use the `NavigationStackHeader`, just as the `StackHeader` uses the `BaseHeader`.

mvaivre (approved):
👏

nop33:
haha, sorry, I removed you as reviewer cause I forgot to fix the pin screen, now it's fixed, please check again 😅

mvaivre (approved):
LGTM :)

PR #226, opened 2023-09-20
Send/Receive buttons, SDK, sorting assets, tx rows, NFTs display

nop33:
Fixes for things discussed yesterday, see commit list


mvaivre:
Awesome. May I ask you something more while you're touching at this? 

image

Could you make sure that the button's text is using the same color as the card's text?
Also, could you align the "New address" button to the right, even when no other buttons are next to it? 

Thanks a lot buddy

nop33:
@mvaivre everything fixed!

image
image
image

mvaivre (approved):
Awesome, great work!

PR #232, opened 2023-09-21
Splashscreen simplification

mvaivre:
Working around react native & expo bugs...


src/screens/DashboardScreen.tsx
```diff
@@ -62,7 +62,7 @@ const DashboardScreen = ({ navigation, ...props }: ScreenProps) => {
   return (
     <DashboardScreenStyled
       refreshControl={
-        <RefreshSpinner refreshing={isLoading} onRefresh={refreshData} progressViewOffset={headerHeight} />
+        <RefreshSpinner refreshing={isLoading} onRefresh={refreshData} progressViewOffset={headerHeight || 170} />
```
mvaivre, line 65: Quick and dirty, but I had to add this because useHeaderHeight doesn't seem to work in the DashboardScreen anymore.

src/components/buttons/HighlightButton.tsx
```diff
@@ -42,7 +42,7 @@ const HighlightButton = ({ title, wide, ...props }: HighlightButtonProps) => {
       <Button
         title={title}
         style={{ backgroundColor: theme.global.accent }}
-        color={theme.font.primary}
+        color="white"
```
mvaivre, line 45: I want to make sure to have max contrast for the highlight button, hence the constant.

android/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
   <uses-permission android:name="android.permission.CAMERA"/>
   <uses-permission android:name="android.permission.INTERNET"/>
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+  <uses-permission android:name="android.permission.RECORD_AUDIO"/>
```
nop33, line 5: This is not really needed, right? I wonder why it was added. I bet you didn't add it manually.

src/App.tsx
```diff
@@ -266,7 +266,9 @@ const Main = ({ children }: { children: ReactNode }) => {
 
   return (
     <RootSiblingParent>
-      <SafeAreaProvider>{children}</SafeAreaProvider>
+      <SafeAreaProvider {...props} style={{ backgroundColor: 'black' }}>
```
nop33, line 269: To make sure that the props.styles does not get overriden by `style={{ backgroundColor: 'black' }}>` we need to do something like:

```tsx
const Main = ({ children, styles, ...props }: ViewProps) => {
...
      <SafeAreaProvider {...props} style={[{ backgroundColor: 'black' }, styles]}>
```

nop33:
Feel free to rebase instead of merging master if you want, it's a small PR.

android/app/src/main/AndroidManifest.xml
mvaivre: I didn't, indeed.

src/App.tsx
mvaivre: It's on purpose. I want to make sure that the background is black.

nop33 (approved):

Issue #215, opened 2023-09-18
Ensure that the user can access the whole content of the page (with scrolling) when the keyboard is visible


nop33:
(for both iOS and Android). Might help: https://www.netguru.com/blog/avoid-keyboard-react-native

- [x] `NewWalletNameScreen`
- [x] `ImportWalletSeedScreen`
- [x] `EditAddressModal`
- [x] `NewAddressScreen`
- [x] `NewContactScreen`
- [x] `EditContactScreen`
- [x] `DestinationScreen`
- [x] `AssetsScreen`
- [x] `SwitchNetworkModal` (custom settings)

nop33:
I propose that we move modals that have input fields into their own screens. It's quite tricky to make them keyboard-aware.

nop33:
Closing since PR was merged.

PR #234, opened 2023-09-21
Keyboard fixes

mvaivre:
Not everything is absolutely perfect, but I feel like it's good enough for a first iteration. Looking forward to your own feeling about this.


src/hooks/layout/useKeyboardMetrics.tsx
nop33, line 1: This was not used after all, right? Wanna keep it?

mvaivre: This could be useful in a second iteration if we decide to go for the more fine grained implementation I told you about on the phone. You're okay with keeping it or you'd rather remove it?

nop33:
It seems that I can't build using the iOS simulator, trying to upgrade everything so that I can test this PR...

nop33 (commented):
I noticed this issue in the edit address modal


https://github.com/alephium/mobile-wallet/assets/1579899/df024145-ab59-4100-9345-df40985bd0e3

nop33:
I would still recommend to move the edit address modal into its own screen, as I proposed here: https://github.com/alephium/mobile-wallet/issues/215#issuecomment-1729893548

android/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
   <uses-permission android:name="android.permission.CAMERA"/>
   <uses-permission android:name="android.permission.INTERNET"/>
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+  <uses-permission android:name="android.permission.RECORD_AUDIO"/>
```
nop33, line 5: ouf, this again, but whatever, seems to be a default, let's leave it.

nop33 (approved):

PR #235, opened 2023-09-21
Fixing empty wallet dashboard, margins, sticky footer

nop33:


mvaivre (approved):
Looks great! I also added a few tweaks on top.

PR #236, opened 2023-09-22
Fix android build

mvaivre:


package.json
```diff
@@ -84,7 +84,8 @@
     "react-qr-code": "^2.0.7",
     "react-redux": "^8.0.1",
     "styled-components": "^5.3.5",
-    "victory-native": "^36.6.10"
+    "victory-native": "^36.6.10",
+    "expo-splash-screen": "~0.20.5"
```
nop33, line 88: This was automatically added I suppose. But we don't need it, right?

mvaivre: Yep.. removed. I hope it doesn't come back 🌬️

nop33 (commented):
I am not so sure we should commit the files in the `android/app/.cxx` directory... Many seem to be just debug files... based on this answer, they shouldn't be committed: https://stackoverflow.com/questions/66129837/about-the-gitignore-question-cxx-folder-in-the-android-project

I also see that `.cxx` is listed here: https://github.com/github/gitignore/blob/main/Android.gitignore

mvaivre:
Closed in favour of #243

Issue #239, opened 2023-09-23
Pending TX polling is heavy and makes the whole apps behave weirdly


mvaivre:
We should probably simplify the process.

Right now, the loading triggers many spinners and UI jumps in all major screens (overview, transfers and addresses).
If we can't find an easy solution, we could :

1) Show a bottom modal when the TX is sent, with a "success animation" and a highlighted "See in explorer" button allowing the user to reach the explorer to check the TX status.
2) And let the user refresh the TX list manually.

This would be acceptable for me as a first step. WDYT?

PR #244, opened 2023-09-25
Fetch address data only when tx becomes confirmed

nop33:
Closes https://github.com/alephium/mobile-wallet/issues/239
Closes https://github.com/alephium/mobile-wallet/issues/245


src/api/addresses.ts
```diff
@@ -62,13 +62,11 @@ export const fetchAddressesTransactions = async (
   const results = []
 
   for (const addressHash of addressHashes) {
-    const txNumber = await client.explorer.addresses.getAddressesAddressTotalTransactions(addressHash)
```
nop33, line 65: I removed this extra call since we don't use this anywhere.

nop33:
@mvaivre This should improve the experience, could you please confirm that this fix addresses https://github.com/alephium/mobile-wallet/issues/239 ?

Thanks 🙏

mvaivre:
Just tested it: it's so much better 👏 Great job! 👏

mvaivre:
image

Few details I noticed:

1) Let's add some margin between the 2. Let's make the area with proposed words standalone, with round corners and reduce its height.
2) The animated button has a weird diamond shape
3) The bottom margin is too big

mvaivre:
I made some changes to the import seed screen layout, please check it out!

There's just the scan button shape left to fix.

src/components/buttons/Button.tsx
```diff
@@ -28,7 +29,7 @@ import { BORDER_RADIUS } from '~/style/globalStyle'
 export interface ButtonProps extends PressableProps {
   title?: string
   type?: 'primary' | 'secondary' | 'transparent' | 'tint'
-  variant?: 'default' | 'contrast' | 'accent' | 'valid' | 'alert'
+  variant?: 'default' | 'contrast' | 'accent' | 'valid' | 'alert' | 'highlight'
```
mvaivre, line 32: Good initiative! But with this new prop we should maybe open an issue to one day get rid of the standalone `HighlightButton` which may be confusing... WDYT?

mvaivre (commented):
Code looks good to me!

src/components/buttons/Button.tsx
nop33: Agreed

nop33: https://github.com/alephium/mobile-wallet/issues/253

mvaivre (approved):
Annnnd, this LGTM!

(PS: while testing thoroughly, I noticed a couple of more things but not related to this PR, and created related issues)

PR #264, opened 2023-09-28
Single wallet

nop33:
Massive code simplification (just look at the LoC diff). Tested quickly with iOS and Android emulators, needs further testing.


ios/Podfile.properties.json
```diff
@@ -1,5 +1,6 @@
 {
   "expo.jsEngine": "hermes",
   "EX_DEV_CLIENT_NETWORK_INSPECTOR": "true",
-  "ios.deploymentTarget": "13.0"
+  "ios.deploymentTarget": "13.0",
+  "newArchEnabled": "false"
```
nop33, line 5: This was added after I was playing around with `prebuild`, I decided to keep this change and to remove others. It doesn't hurt since it reflects what we have in the app config file.

package.json
```diff
@@ -36,7 +36,7 @@
     "buffer": "^6.0.3",
     "colord": "^2.9.3",
     "dayjs": "^1.11.2",
-    "expo": "49.0.7",
+    "expo": "49.0.13",
```
nop33, line 39: I updated expo and secure-store just to make sure that our Face ID problems are not because of old versions.

src/persistent-storage/settings.ts
```diff
@@ -62,7 +64,7 @@ const constructSettingsStorageKey = (key: SettingsKey) => `${STORAGE_KEY}-${key}
 
 export const loadSettings = async (key: SettingsKey): Promise<SettingsPartial> => {
   try {
-    const rawSettings = await AsyncStorage.getItem(constructSettingsStorageKey(key))
+    const rawSettings = await SecureStore.getItemAsync(constructSettingsStorageKey(key))
```
nop33, line 67: To keep the code simple, I added the `useBiometrics` in the general settings and moved the whole general settings into the SecureStore. We still have plently of space before we reach the 2048 byte limit, so instead of separating the biometrics settings storage into its own entry, I just moved it all into SecureStoe.

mvaivre (approved):

PR #272, opened 2023-10-02
Backup mnemonic flow and UI fixes

nop33:


src/screens/BackupMnemonic/BackupIntroScreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+              {
+                text: 'In the following screens you will see and verify your secret recover phrase.',
+                type: 'secondary'
+              },
+              { text: 'Write it down and store it in a safe place.', type: 'primary' },
+              {
+                text: 'Why is this important?',
+                type: 'link',
+                // TODO: Put correct link
+                url: 'https://docs.alephium.org/Frequently-Asked-Questions.html'
```
nop33, line 61: We need an FAQ entry

src/App.tsx
```diff
@@ -227,6 +228,10 @@ const Main = ({ children, ...props }: ViewProps) => {
         Alert.alert('Authentication required', 'Please authenticate to unlock your wallet.', [
           { text: 'Try again', onPress: unlockApp }
         ])
+      } else if (error.message?.includes('Could not decrypt the value with provided keychain')) {
+        console.error(e)
+        await deleteWallet()
+        navigateRootStack('LandingScreen')
```
nop33, line 234: It seems that when uninstalling the app, SecureStore leaves some "garbage" behind, so that next time the app gets installed, we get this error.

src/App.tsx
```diff
@@ -237,8 +242,7 @@ const Main = ({ children, ...props }: ViewProps) => {
 
   useEffect(() => {
     // We want this to only run 1 time and not every time lastNavigationState changes (dep of unlockApp).
-    // On Android the AppState.change event is triggered when launching the app, but not on iOS
```
nop33, line 240: Seems like this happens also in Android.

src/screens/LandingScreen.tsx
```diff
@@ -86,8 +88,14 @@ const LandingScreen = ({ navigation, ...props }: LandingScreenProps) => {
...
 
+  const handleScreenLayoutChange = (e: LayoutChangeEvent) => {
+    const { width, height } = e.nativeEvent.layout
+
+    setDimensions({ width, height })
+  }
+
   return (
-    <Screen contrastedBg {...props}>
+    <Screen contrastedBg {...props} onLayout={handleScreenLayoutChange}>
```
nop33, line 98: On Android the height from the `Dimensions` was not the same as the one of the Screen. This fixes it.

src/screens/SendReceive/Send/AssetsScreen/AssetRow.tsx
```diff
@@ -91,23 +90,16 @@ const AssetRow = ({ asset, style, isLast }: AssetRowProps) => {
     setIsSelected(isNowSelected)
 
     if (isNowSelected) {
-      inputRef.current?.focus()
+      setTimeout(() => inputRef.current?.focus(), 500)
```
nop33, line 93: I spent a lot time on this... This is the hack I found so that the focus happens after the animation finishes. I suspect that it's due to the animation that the keyboard was not showing. This "fix" seems to work.

src/screens/SendReceive/Send/AssetsScreen/AssetRow.tsx
```diff
@@ -160,13 +152,14 @@ const AssetRow = ({ asset, style, isLast }: AssetRowProps) => {
                 onChangeText={handleOnAmountChange}
                 keyboardType="number-pad"
                 inputMode="decimal"
-                onBlur={handleInputBlur}
                 ref={inputRef}
+                multiline={true}
+                numberOfLines={2}
```
nop33, line 157: When choosing "Use max" many times we get a very long number. This solution should be good enough to cover all cases.

src/App.tsx
mvaivre: Couldn't that be dangerous? Are we sure that this error can only be shown because of this "garbage" left behind, and not because of some user action?

mvaivre:
Start of the functional tests:

If I "fail" the FaceID scan before showing the mnemonic on validation, I get this:
IMG_1315

... and then the backdrop stay mounted and I can't interact with the screen anymore.

src/App.tsx
nop33: I think I fixed the problem without this problematic fix. I set the `keychainService` option of the SecureStore `setItemAsync` (as well as `getItemAsync`) to the string `'mobile-wallet-key-service'`. This seems to not cause this exception. The way I tested it was:

- Have the app running on my phone
- Stop the server
- Change the app name
- Start the server
- I see message on my phone that says "App needs to be closed while being updated"
- After a few seconds the app launches again, it has the new name, and no error was shown (tested this with both pin and bio login, all worked 💪 )

nop33:
> If I "fail" the FaceID scan before showing the mnemonic on validation, I get this:

Ah, good catch! We are catching this error in all other places where we use authentication, but I forgot to add it also to this one!

app.config.js
```diff
@@ -36,14 +36,15 @@ export default {
...
+      bundleIdentifier: IS_DEV ? 'org.alephium.mobilewallet.dev' : 'org.alephium.mobilewallet'
     },
     android: {
       adaptiveIcon: {
         foregroundImage: './assets/adaptive-icon.png',
         backgroundColor: '#000000'
       },
-      package: IS_DEV ? 'org.alephium.wallet.dev' : 'org.alephium.wallet'
+      package: IS_DEV ? 'org.alephium.wallet.dev' : 'org.alephium.wallet',
+      versionCode: 2
```
nop33, line 47: Every time we submit a build we need to iterate this number.

app.config.js
```diff
@@ -36,14 +36,15 @@ export default {
     assetBundlePatterns: ['**/*'],
     ios: {
       supportsTablet: true,
-      bundleIdentifier: IS_DEV ? 'org.alephium.wallet.dev' : 'org.alephium.wallet'
+      bundleIdentifier: IS_DEV ? 'org.alephium.mobilewallet.dev' : 'org.alephium.mobilewallet'
```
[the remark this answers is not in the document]
mvaivre: Mhh what? The error message is super weird "An App ID with Identified 'org.alephium.org' is
not available. Please enter a different string." 🤔 

But anyway, using mobilewallet is fine.

src/screens/Addresses/Contact/ContactFormBaseScreen.tsx
```diff
@@ -18,33 +18,41 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 
   return (
-    <>
+    <ScrollScreen
+      usesKeyboard
+      fill
+      hasNavigationHeader
+      headerOptions={{
+        type: 'stack',
+        headerRight: () => <ContinueButton title={buttonText} onPress={handleSubmit(onSubmit)} />,
```
mvaivre, line 51: We should use another icon than the right arrow for this one, as there's no step after saving. A checkmark could be better IMO. What do you think?

src/components/layout/ScrollScreen.tsx
```diff
@@ -82,6 +82,7 @@ const ScrollScreen = ({
         onScroll={hasNavigationHeader ? navigationScrollHandler : screenScrollHandler}
         onScrollEndDrag={scrollEndHandler}
         showsHorizontalScrollIndicator={false}
+        keyboardShouldPersistTaps="always"
```
mvaivre, line 85: Mhh are we sure we want this to be true for all ScrollScreen? 

It means that tapping on a ScrollScreen will never dismiss the keyboard. Is there a specific reason to add this here instead of on a "per screen basis"?

mvaivre: This is indeed an issue: try generating a new address and choosing a group :)

src/screens/Addresses/Contact/ContactFormBaseScreen.tsx
mvaivre: Also true for the address edit screen and the new address screen

src/screens/DashboardScreen.tsx
```diff
@@ -119,6 +162,47 @@ const DashboardScreen = ({ navigation, ...props }: ScreenProps) => {
...
       )}
+      <Portal>
+        <BottomModal
+          isOpen={isBackupReminderModalOpen}
+          onClose={() => setIsBackupReminderModalOpen(false)}
+          Content={(props) => (
+            <ModalContent verticalGap {...props}>
+              <ScreenSection>
+                <BottomModalScreenTitle>
+                  {isNewWallet ? 'Welcome to your new wallet! 👋' : "Let's verify! 😌"}
```
mvaivre, line 173: Nit: Maybe keeping the title on one line could be nicer :) For the first one, what about saying something like: Welcome! 👋  or Warm welcome! 👋 ?

mvaivre:
Edge case:

IMG_1317

First of all: crazy that the delta in worth is 0 after a month? It doesn't look like it's the case when looking at the graph 🤔

If it's the case, we may use a grey color for the worth delta and the graph IMO. WDYT?

src/screens/SendReceive/Send/AssetsScreen/AssetRow.tsx
mvaivre: nice one.

mvaivre (commented):
Awesome work 👏 👏

nop33:
> First of all: crazy that the delta in worth is 0 after a month? It doesn't look like it's the case when looking at the graph 🤔

> If it's the case, we may use a grey color for the worth delta and the graph IMO. WDYT?

Delta is fixed now, it was a mistake on my side!

src/screens/Addresses/Contact/ContactFormBaseScreen.tsx
nop33: Sure! I'm on it!

src/components/layout/ScrollScreen.tsx
nop33: Ah, I see, it was solving a few issues in the Assets screen. I'll make sure to move it there specifically! Thanks for finding this!

src/screens/DashboardScreen.tsx
nop33: "A warm welcome 👋"? https://englishgrammarmaster.quora.com/Which-is-correct-a-warm-welcome-or-warm-welcome

BTW, we already say "Welcome to Alephium! Enjoy your new wallet!" in the previous screen. Maybe we could say something simpler in the modal, like:

"Hello there 👋"

WDYT?

mvaivre: Yes! I like it.

src/components/headers/BaseHeader.tsx
```diff
@@ -139,8 +145,8 @@ const BaseHeader = ({
   const compactContentAnimatedStyle = useAnimatedStyle(() =>
     hasCompactHeader
       ? {
-          opacity: interpolate(scrollY?.value || 0, [60, scrollEndThreshold], [0, 1], Extrapolate.CLAMP),
-          height: interpolate(scrollY?.value || 0, [60, scrollEndThreshold], [110, 95], Extrapolate.CLAMP)
+          opacity: interpolate(scrollY?.value || 0, [40, scrollEndThreshold], [0, 1], Extrapolate.CLAMP),
```
mvaivre, line 148: NIT: Why change the initial interp value to 40 instead on 60? On iOS it makes the animation less smooth (we see the compact header being static before moving up).

src/components/layout/ScrollScreen.tsx
mvaivre: Is it going to be fixed in the upcoming PR then?

mvaivre (approved):
Just tiny comments 🚀 I pre-approve.

src/components/layout/ScrollScreen.tsx
nop33: Correct!

src/screens/DashboardScreen.tsx
nop33: Next PR 😇

src/components/headers/BaseHeader.tsx
nop33: Ah shit, I wanted to do this only on Android! Good catch! I'll add it in my next PR

PR #297, opened 2023-10-05
Edit wallet name

### LINKED ISSUE #132: Change wallet name
Since we don't allow wallets having the same name (#122), we should allow changing the name of a wallet.

nop33:
Closes #132

### DIFF
--- a/src/navigation/RootStackNavigation.tsx
+++ b/src/navigation/RootStackNavigation.tsx
@@ -51,6 +51,7 @@ import NewWalletNameScreen from '~/screens/new-wallet/NewWalletNameScreen'
 import NewWalletSuccessScreen from '~/screens/new-wallet/NewWalletSuccessScreen'
 import PinCodeCreationScreen from '~/screens/new-wallet/PinCodeCreationScreen'
 import SelectImportMethodScreen from '~/screens/new-wallet/SelectImportMethodScreen'
+import EditWalletNameScreen from '~/screens/Settings/EditWalletName'
 import SettingsScreen from '~/screens/Settings/SettingsScreen'
 import SplashScreen from '~/screens/SplashScreen'
 import { routeChanged } from '~/store/appSlice'
@@ -171,6 +172,7 @@ const RootStackNavigation = () => {
                     <RootStack.Screen name="DecryptScannedMnemonicScreen" component={DecryptScannedMnemonicScreen} />
                     <RootStack.Screen name="ImportWalletSeedScreen" component={ImportWalletSeedScreen} />
                     <RootStack.Screen name="AddBiometricsScreen" component={AddBiometricsScreen} />
+                    <RootStack.Screen name="EditWalletNameScreen" component={EditWalletNameScreen} />
                     <RootStack.Screen
                       name="ImportWalletAddressDiscoveryScreen"
                       component={ImportWalletAddressDiscoveryScreen}
--- a/src/navigation/rootStackRoutes.ts
+++ b/src/navigation/rootStackRoutes.ts
@@ -53,6 +53,7 @@ type RootStackParamList = {
   }
   SelectImportMethodScreen: undefined
   DecryptScannedMnemonicScreen: undefined
+  EditWalletNameScreen: undefined
 }
 
 export default RootStackParamList
--- a/src/screens/Settings/EditWalletName.tsx
+++ b/src/screens/Settings/EditWalletName.tsx
@@ -1,0 +1,89 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { getHumanReadableError } from '@alephium/sdk'
+import { StackScreenProps } from '@react-navigation/stack'
+import { usePostHog } from 'posthog-react-native'
+import { useState } from 'react'
+import Toast from 'react-native-root-toast'
+
+import { ContinueButton } from '~/components/buttons/Button'
+import Input from '~/components/inputs/Input'
+import { ScreenSection } from '~/components/layout/Screen'
+import ScrollScreen, { ScrollScreenProps } from '~/components/layout/ScrollScreen'
+import SpinnerModal from '~/components/SpinnerModal'
+import { useAppDispatch, useAppSelector } from '~/hooks/redux'
+import RootStackParamList from '~/navigation/rootStackRoutes'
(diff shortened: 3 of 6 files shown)


src/App.tsx
```diff
@@ -251,10 +251,10 @@ const Main = ({ children, ...props }: ViewProps) => {
 
   useEffect(() => {
     const handleAppStateChange = (nextAppState: AppStateStatus) => {
-      if (nextAppState === 'background' && !isCameraOpen) {
+      if (nextAppState === 'background' && walletMnemonic && !isCameraOpen) {
```
nop33, line 254: We only trigger the "appBecameInactive" when we actually have a mnemonic (meaning, a wallet). Otherwise, it means we don't have one yet, so we are in the wallet creation flow.

src/App.tsx
```diff
@@ -251,10 +251,10 @@ const Main = ({ children, ...props }: ViewProps) => {
...
         navigateRootStack('SplashScreen')
         dispatch(appBecameInactive())
-      } else if (nextAppState === 'active' && !walletMnemonic && !isUnlockingApp && !isCameraOpen) {
+      } else if (nextAppState === 'active' && !walletMnemonic && !isCameraOpen && !isUnlockingApp) {
```
nop33, line 257: Just a reordering to match the order of the condition above 😇

src/components/AuthenticationModal.tsx
```diff
@@ -48,22 +49,28 @@ const errorInstructionSet: Instruction[] = [
...
     try {
-      const storedWallet = await getStoredWallet(usePin)
-      const usesBiometrics = usePin ? false : await loadBiometricsSettings()
+      const storedWallet = await getStoredWallet(forcePinUsage)
+
+      // This should never happen, but if it does, inform the user instead of being stuck
+      if (!storedWallet) {
+        Alert.alert('Missing wallet', 'Could not find wallet to authenticate. Please, restart the app')
+        return
+      }
```
nop33, line 66: This component should only be used when we have a wallet and we want to authenticate. So, in the event that this is used without a wallet (which should be never), we should still catch it and make sure the user doesn't get stuck.

src/components/AuthenticationModal.tsx
```diff
@@ -113,12 +117,16 @@ const ConfirmWithAuthModal = ({ onConfirm, onClose, usePin = false }: ConfirmWit
           <CenteredInstructions instructions={shownInstructions} />
           <PinCodeInput pinLength={pinLength} onPinEntered={decryptMnemonic} />
         </ModalContent>
+      ) : (
+        <Message>
+          <AppText>Loading wallet...</AppText>
+        </Message>
```
nop33, line 123: To make sure there is never just a white screen. Should only be displayed while the fingerprint sensor is visible.

src/components/ModalWithBackdrop.tsx
```diff
@@ -19,13 +19,13 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 
-interface ModalWithBackdropProps extends ModalProps {
+export interface ModalWithBackdropProps extends ModalProps {
   closeModal?: () => void
   color?: string
 }
 
 const ModalWithBackdrop = ({ children, closeModal, color, ...props }: ModalWithBackdropProps) => (
-  <Modal transparent={true} {...props} animationType="none">
+  <Modal transparent={true} animationType="none" {...props}>
```
nop33, line 28: so that `animationType` can be overridden if needed.

src/screens/LoginWithPinScreen.tsx
```diff
@@ -58,9 +60,20 @@ const LoginWithPinScreen = (props: LoginWithPinScreenProps) => {
 
   return (
     <Screen contrastedBg {...props}>
-      {isPinModalVisible && <ConfirmWithAuthModal usePin onConfirm={handleSuccessfulLogin} />}
+      <AuthenticationModal visible={isPinModalVisible} forcePinUsage onConfirm={handleSuccessfulLogin} />
+      {!isPinModalVisible && (
+        <Message>
+          <AppText>Nice 👍 Unlocking...</AppText>
+        </Message>
+      )}
```
nop33, line 68: This will be shown while we are loading some initial data when the app is first launched. Then it will disappear much quicker after data has been initialized.

src/components/ModalWithBackdrop.tsx
mvaivre: Yeah I forced it to none because these animations posed problems in the past. Let's just keep that in mind in case we see witness behaviours in the future :)

src/screens/LoginWithPinScreen.tsx
mvaivre: What about showing the loading spinner instead? Just a bit less confusing IMO as you don't have to read anything, you understand that you just need to wait.

mvaivre (commented):
🚂 👏

src/components/ModalWithBackdrop.tsx
nop33: It's good to default to none, but be able to override it if needed though, don't you agree?

src/screens/LoginWithPinScreen.tsx
nop33: Sure. My initial thought is that by having some text it'd be easier to debug in case someone gets stuck on this screen for whatever reason. I'll leave the word "Unlocking" with a spinner, if you don't mind.

src/components/ModalWithBackdrop.tsx
nop33: We are currently using `"fade"` in these places:
- Spinner modal
- Consolidation modal
- Color picker modal
- Verify Mnemonic modal
- Settings screen

We should check each individual case on iOS to make sure we don't get stuck anywhere.

src/screens/LoginWithPinScreen.tsx
nop33: Done ✅

src/screens/AddressDiscoveryScreen.tsx
```diff
@@ -67,6 +68,9 @@ const AddressDiscoveryScreen = ({ navigation, route: { params }, ...props }: Scr
 
   const stopScan = () => dispatch(addressDiscoveryStopped())
 
+  const continueToNextScreen = () =>
+    resetNavigationState(isImporting ? 'NewWalletSuccessScreen' : 'InWalletTabsNavigation')
```
nop33, line 72: Resetting the navigation state helps with preventing going back to the wallet creation procedure. There's no point doing that since the wallet has already been created at this point.

nop33:
I rebased it on `internal-feedback-polish-2` (after all the work we did today with the `/android` and `/ios` folders) and also added my "custom network" changes so that we only have 1 PR open each.

Tomorrow morning we can start by merging yours, then reviewing, testing and merging this one.

@mvaivre, I would focus the functional tests on your iPhone device and mainly to try all functions that show a spinner modal: https://github.com/alephium/mobile-wallet/pull/297#discussion_r1349969208

src/components/AddressCard.tsx
```diff
@@ -124,8 +124,9 @@ const AddressCard = ({ style, addressHash, onSettingsPress }: AddressCardProps)
           shadowOffset: { height: 5, width: 0 },
           shadowOpacity: theme.name === 'dark' ? 0.5 : 0.15,
           shadowRadius: 8,
-          elevation: 10,
-          borderColor: outterBorderColor
+          elevation: 3,
+          borderColor: outterBorderColor,
+          overflow: 'hidden'
```
nop33, line 129: I had to change the elevation and also add the overflow, cause it didn't look as good on Android as it did on iOS:

image

mvaivre (approved):

PR #300, opened 2023-10-05
Internal feedback polish 2

mvaivre:


nop33 (approved):

src/components/AddressCard.tsx
```diff
@@ -81,81 +95,147 @@ const AddressCard = ({ style, addressHash, onSettingsPress }: AddressCardProps)
...
+              customIcon={
+                <DefaultAddressBadge
+                  strokeOnly={!isDefaultAddress}
+                  size={18}
+                  color={isDefaultAddress ? theme.global.accent : textColor}
+                />
+              }
+              style={{ backgroundColor: isDefaultAddress ? 'rgba(255, 255, 255, 0.2)' : buttonsBackground }}
+              round
+            />
```
nop33, line 161: If the address is already the default one, we don't need to call the `handleDefaultAddressToggle`. It makes an unnecessary call to the store, displays the spinner modal, and the toast. I think it should only do all that when changing the default address.

nop33 (approved):

PR #305, opened 2023-10-09
Change network settings with badge tap

### LINKED ISSUE #302: Allow network switching by tapping on network badge
This was requested by Thiago, and it's something I also had in mind. Tapping on the network badge can open the bottom modal with the 3 network options.

With this, I would also like to take the opportunity to move the custom network settings to their own screen, instead of having them inside the bottom modal. The reason is that the scrollable modal with an active keyboard is still not working very well.

nop33:
Closes #302


nop33:
Closing for PR consolidation in favor of this https://github.com/alephium/mobile-wallet/pull/297

PR #317, opened 2023-10-16
Fix app unlock handling

nop33:


src/navigation/RootStackNavigation.tsx
```diff
@@ -84,104 +98,193 @@ const RootStackNavigation = () => {
...
-                    <RootStack.Screen
-                      name="ImportWalletAddressDiscoveryScreen"
-                      component={ImportWalletAddressDiscoveryScreen}
-                    />
-                  </RootStack.Group>
-                </RootStack.Navigator>
-              </WalletConnectContextProvider>
-            </AnalyticsProvider>
-          </NavigationScrollContextProvider>
+          <AppUnlockHandler>
```
nop33, line 101: I creating the `<AppUnlockHandler>` component right after the `<NavigationContainer>` one so that I have access to the `useNavigation` hook. This way, we don't need the customized `navigateRootStack` function and we are sure that the navigator is always ready.

src/navigation/RootStackNavigation.tsx
```diff
@@ -84,104 +98,193 @@ const RootStackNavigation = () => {
...
+          </AppUnlockHandler>
         </NavigationContainer>
       </Host>
     </GestureHandlerRootView>
   )
 }
 
 export default RootStackNavigation
+
+const AppUnlockHandler = ({ children }: { children: ReactNode }) => {
```
nop33, line 205: This code is simply copied from the `App.tsx`, with the only difference that I am now using `navigation.navigate` instead of the custom `navigateRootStack` function.

src/navigation/RootStackNavigation.tsx
```diff
@@ -84,104 +98,193 @@ const RootStackNavigation = () => {
...
+      } else if (nextAppState === 'active' && !walletMnemonic && !isCameraOpen) {
+        unlockApp()
+      }
+
+      appState.current = nextAppState
+    }
+
+    if (!isAppStateChangeCallbackRegistered) {
+      handleAppStateChange(appState.current)
+    }
```
nop33, line 280: When the app is initially launched (either after a fresh install, a device restart, or if the user quits the app and launches it again), the `'change'` event of `AppState` does not get triggered (kinda makes sense). Thus, we manually need to run the code that handles the event, only for the first time.

src/utils/navigation.ts
```diff
@@ -50,15 +49,3 @@ export const setNavigationState = (state: NavigationState) =>
...
-// https://reactnavigation.org/docs/navigating-without-navigation-prop
-export const navigateRootStack = <RouteName extends keyof RootStackParamList & keyof SendNavigationParamList>(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  name: any, // TODO: Figure out why we can't use `RouteName` instead
-  params?: RootStackParamList[RouteName] | SendNavigationParamList[RouteName]
-) => {
-  if (rootStackNavigationRef.isReady()) {
-    rootStackNavigationRef.navigate(name, params)
-  }
-}
```
nop33, line 64: So happy to get rid of this.

src/navigation/RootStackNavigation.tsx
mvaivre: That's great!

mvaivre (approved):
LFG

PR #335, opened 2023-10-18
Post beta release fixes

nop33:
## WalletConnect deep link

Seems like this was an easy fix! I didn't spend much time on it, just followed the guide from:
- https://docs.expo.dev/guides/linking/#linking-to-your-app
- https://docs.expo.dev/versions/latest/sdk/linking/

It works well when tested on my Android phone with the dev client. I've tested using the following different methods:
- Opening a WalletConnect URL through a mobile browser (trying to browse it)
- Going on a production Alephium dApp (example: walph.io), clicking _Connect Alephium_ -> _WalletConenct_ -> _Connect_
- Entering this in the console:
   ```
   npx uri-scheme open "wc:[WALLETCONNECT_URI_HERE]" --android
   ```

All the above worked ✅ 

It works when the app is active, but also when it's inactive.

I am curious to see if it also works on an iOS device...

(You see a red logo cause that's my dev client)


mvaivre (approved):
👏 👏 
LGTM

Issue #331, opened 2023-10-18
Chart disappears when selecting 1D option


nop33:
An iOS tester reported through the TestFlight report system:
> App is nice. Noticed a bug where the returns graph disappears when the 1D option is selected. Cannot switch timeframes once changed to 1D
>
> ![image](https://github.com/alephium/mobile-wallet/assets/1579899/974b0174-990b-41f3-bafc-474d42efef1f)

Device info

<img width="685" alt="image" src="https://github.com/alephium/mobile-wallet/assets/1579899/ff080481-27da-4b14-9f2b-618ed5d9307e">

nop33:
Another person reported this as well.

nop33:
I've asked the community to verify whether this bug was fixed with the latest update or not.

nop33:
After the last update the user who reported it on Discord could not reproduce it. I'll close it for now, let's see if it re-appears!

Issue #340, opened 2023-10-19
Display group in address boxes


nop33:
> I wanted to transfer to my address in group 0, to interact with dapps. I simple don't know which one it is
> I see the info on the addresses section 
> But not when clicking on the receive button from the overview (which sends me to the list page)

<img width="534" alt="image" src="https://github.com/alephium/mobile-wallet/assets/1579899/c4a9a51b-e31d-46e7-99f3-3052910a8e01">

nop33:
Another person asked for this feature yesterday

PR #343, opened 2023-10-19
Post beta fixes

nop33:
Possibly related to #331

### DIFF
--- a/src/components/AddressBox.tsx
+++ b/src/components/AddressBox.tsx
@@ -23,11 +23,16 @@ import styled, { useTheme } from 'styled-components/native'
 
 import { fastestSpringConfiguration } from '~/animations/reanimated/reanimatedAnimations'
 import AddressBadge from '~/components/AddressBadge'
+import AppText from '~/components/AppText'
 import AssetAmountWithLogo from '~/components/AssetAmountWithLogo'
 import Checkmark from '~/components/Checkmark'
 import { useAppSelector } from '~/hooks/redux'
-import { makeSelectAddressesKnownFungibleTokens, makeSelectAddressesNFTs } from '~/store/addressesSlice'
-import { BORDER_RADIUS, VERTICAL_GAP } from '~/style/globalStyle'
+import {
+  makeSelectAddressesKnownFungibleTokens,
+  makeSelectAddressesNFTs,
+  selectAddressByHash
+} from '~/store/addressesSlice'
+import { BORDER_RADIUS, DEFAULT_MARGIN, VERTICAL_GAP } from '~/style/globalStyle'
 import { AddressHash } from '~/types/addresses'
 import { ImpactStyle, vibrate } from '~/utils/haptics'
 
@@ -39,6 +44,7 @@ interface AddressBoxProps extends PressableProps {
 const AnimatedPressable = Animated.createAnimatedComponent(Pressable)
 
 const AddressBox = ({ addressHash, isSelected, onPress, ...props }: AddressBoxProps) => {
+  const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
   const selectAddressesKnownFungibleTokens = useMemo(makeSelectAddressesKnownFungibleTokens, [])
   const knownFungibleTokens = useAppSelector((s) => selectAddressesKnownFungibleTokens(s, addressHash))
   const selectAddressesNFTs = useMemo(makeSelectAddressesNFTs, [])
@@ -59,7 +65,12 @@ const AddressBox = ({ addressHash, isSelected, onPress, ...props }: AddressBoxPr
     <AddressBoxStyled {...props} onPress={handlePress} style={[boxAnimatedStyle, props.style]}>
       <AddressBoxTop>
         <AddressBadgeStyled onPress={handlePress} addressHash={addressHash} textStyle={{ fontSize: 18 }} />
-        {isSelected && <Checkmark />}
+        <Group>
+          <AppText color="tertiary" size={14}>
+            group {address?.group}
+          </AppText>
+          {isSelected && <Checkmark />}
+        </Group>
       </AddressBoxTop>
       <AddressBoxBottom>
         <AssetsRow>
@@ -99,6 +110,8 @@ const AddressBoxTop = styled.View`
   background-color: ${({ theme }) => theme.bg.highlight};
   border-top-right-radius: 6px;
   border-top-left-radius: 6px;
+  gap: ${DEFAULT_MARGIN}px;
+  align-items: center;
 `
 
 const AddressBoxBottom = styled.View`
@@ -110,10 +123,17 @@ const AddressBoxBottom = styled.View`
 
 const AddressBadgeStyled = styled(AddressBadge)`
   max-width: 80%;
+  flex-shrink: 1;
 `
 
 const AssetsRow = styled.View`
   flex-direction: row;
   flex-wrap: wrap;
   gap: 12px;
+`
+
+const Group = styled.View`
+  flex-direction: row;
+  gap: ${DEFAULT_MARGIN}px;
+  flex-shrink: 0;
 `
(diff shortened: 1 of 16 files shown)


src/navigation/RootStackNavigation.tsx
```diff
@@ -231,7 +235,7 @@ const AppUnlockHandler = ({ children }: { children: ReactNode }) => {
       if (!wallet) {
         if (lastNavigationState) {
           // When we are at the wallet creation flow we want to reset to the last screen
-          setNavigationState(lastNavigationState)
+          restoreNavigation(navigation, lastNavigationState)
```
nop33, line 238: Have a look at my comments on the implementation of this function later in this PR, I explain why I add the `navigation` object as prop.

src/navigation/RootStackNavigation.tsx
```diff
@@ -268,7 +270,7 @@ const AppUnlockHandler = ({ children }: { children: ReactNode }) => {
     const handleAppStateChange = (nextAppState: AppStateStatus) => {
       if (nextAppState === 'background' && walletMnemonic && !isCameraOpen) {
         loadBiometricsSettings().then((isBioEnabled) =>
-          navigation.navigate(isBioEnabled ? 'LandingScreen' : 'LoginWithPinScreen')
+          resetNavigation(navigation, isBioEnabled ? 'LandingScreen' : 'LoginWithPinScreen')
```
nop33, line 273: I had to reset and not navigate, because if the user cancels the bio auth prompt, then they can swipe "back" and go to the previous screen. Resetting fixes it.

src/screens/BackupMnemonic/VerificationSuccessScreen.tsx
```diff
@@ -46,7 +46,7 @@ const VerificationSuccessScreen = ({ navigation, ...props }: VerificationSuccess
       />
     </ScreenSection>
     <ScreenSection>
-      <Button variant="highlight" title="Return to my wallet" onPress={() => resetNavigationState()} />
+      <Button variant="highlight" title="Return to my wallet" onPress={() => resetNavigation(navigation.getParent())} />
```
nop33, line 49: We need to get the parent because we are in a sub-navigation. TypeScript will properly complain if the wrong navigator is passed, which makes our code safer.

src/screens/Settings/SettingsScreen.tsx
```diff
@@ -307,7 +308,9 @@ const SettingsScreen = ({ navigation, ...props }: ScreenProps) => {
           isOpen={isWalletDeleteModalOpen}
           onClose={() => setIsWalletDeleteModalOpen(false)}
           maximisedContent={Platform.OS === 'ios'}
-          Content={(props) => <WalletDeleteModal {...props} />}
+          Content={(props) => (
+            <WalletDeleteModal onDelete={() => resetNavigation(navigation, 'LandingScreen')} {...props} />
+          )}
```
nop33, line 313: This is the best example why I am passing the `navigation` object as a prop to the `resetNavigation` function. The `WalletDeleteModal` does not have access to the `navigation` object (it's not a screen). If we'd use `useNavigate` to get the `navigation` object in the modal, TypeScript wouldn't complain and compilation would pass. If we'd try to mount the component though, we'd get an exception. Always requiring a valid `navigation` object ensures that we'll never get to that state.

src/store/assets/priceApiSlice.ts
```diff
@@ -49,8 +49,23 @@ export const priceApi = createApi({
...
+
+          if (!isDuplicatedItem && itemDate !== today)
+            acc.push({
+              date: itemDate,
+              price
+            })
+
+          return acc
+        }, [] as HistoricalPriceResult[])
+      }
```
nop33, line 68: This is copied directly from the dw

src/utils/navigation.ts
```diff
@@ -44,8 +44,13 @@ export const isNavStateRestorable = (state: NavigationState) => {
...
+}
 
-export const resetNavigationState = (initialRouteName?: string) =>
-  rootStackNavigationRef.resetRoot(getInitialNavigationState(initialRouteName))
+export const resetNavigation = (
+  navigation: NavigationProp<RootStackParamList>,
+  initialRouteName?: keyof RootStackParamList
+) => {
+  navigation.dispatch(CommonActions.reset(getInitialNavigationState(initialRouteName)))
+}
```
nop33, line 56: Finally getting rid of the `rootStackNavigationRef.resetRoot` which were problematic and were causing unexpected issues. Our code is more type-safe now. We always require a valid navigation object before resetting/restoring the navigation.

src/navigation/RootStackNavigation.tsx
```diff
@@ -105,8 +108,9 @@ const RootStackNavigation = () => {
               <AnalyticsProvider>
                 <WalletConnectContextProvider>
                   <RootStack.Navigator initialRouteName="LandingScreen">
-                    {/* Sub-navigation with custom header */}
-                    <RootStack.Group screenOptions={{ headerTransparent: true }}>
+                    {/* Sub-navigation with custom header. Showing the header only when authenticated fixes the
+                    reanimated bug while still allowing animated transitions between screens */}
+                    <RootStack.Group screenOptions={{ headerTransparent: true, headerShown: isAuthenticated }}>
```
nop33, line 113: This is quite funny (and annoying), but after implementing the proper way of resetting the navigation (see [docs](https://reactnavigation.org/docs/7.x/navigation-actions#reset)), I got the same problem that we faced where a reanimated exception would happen when resetting the navigation to a screen that is using the react-navigation header. I figured that if I hide the header when the app is in the background the error doesn't happen. Not sure why when using the old navigation resetting method this problem was not happening.

src/navigation/RootStackNavigation.tsx
```diff
@@ -255,9 +259,7 @@ const AppUnlockHandler = ({ children }: { children: ReactNode }) => {
       const error = e as { message?: string }
 
       if (error.message?.includes('User canceled')) {
-        Alert.alert('Authentication required', 'Please authenticate to unlock your wallet.', [
-          { text: 'Try again', onPress: unlockApp }
-        ])
+        showToast('Authentication required. Exit the app and try again.')
```
nop33, line 262: Showing an alert was creating this problem: https://github.com/alephium/mobile-wallet/issues/344

So I thought we should quit the app if the user cancels authentication. But after researching I found out that there's no cross-platform native way to do this and I also read comments that iOS prevents apps from closing themselves.

The simpler solution to this is to display a toast and ask the user to refresh the app.

src/components/AddressBox.tsx
```diff
@@ -59,7 +65,12 @@ const AddressBox = ({ addressHash, isSelected, onPress, ...props }: AddressBoxPr
     <AddressBoxStyled {...props} onPress={handlePress} style={[boxAnimatedStyle, props.style]}>
       <AddressBoxTop>
         <AddressBadgeStyled onPress={handlePress} addressHash={addressHash} textStyle={{ fontSize: 18 }} />
-        {isSelected && <Checkmark />}
+        <Group>
+          <AppText color="tertiary" size={14}>
+            group {address?.group}
+          </AppText>
+          {isSelected && <Checkmark />}
+        </Group>
```
nop33, line 73: At least 2 people came to me asking for this feature https://github.com/alephium/mobile-wallet/issues/340

Their arguments made a lot of sense to me.

src/utils/navigation.ts
mvaivre: Yeah that feels good!

src/components/AddressBox.tsx
mvaivre: Definitely.

mvaivre: I've just seen that having the Checkmark next to the group screw up the vertical alignement (the group label is too high).

src/navigation/RootStackNavigation.tsx
mvaivre: That's a very weird one, but hey, it works.

mvaivre:
Functional and code review done. Just a tiny comment ⬆️

src/components/AddressBox.tsx
nop33: Is it? I didn't notice that... Fixed ✅

nop33:
Approved through Slack ✅

PR #352, opened 2023-10-23
Post beta fixes

nop33:
@mvaivre I am not sure why we have so few analytics recorded on PostHog... My dev client seems to be working fine once I remove the condition that checks if `__DEV__` is true.

https://stackoverflow.com/questions/34315274/react-native-detect-dev-or-production-env

I hope this PR fixes the issue...

### DIFF
--- a/src/components/AddressCard.tsx
+++ b/src/components/AddressCard.tsx
@@ -59,7 +59,7 @@ const AddressCard = ({ style, addressHash, onSettingsPress }: AddressCardProps) 
 
   const [loading, setLoading] = useState(false)
 
-  const totalAddressBalance = BigInt(address?.balance ?? 0) + BigInt(address?.lockedBalance ?? 0)
+  const totalAddressBalance = BigInt(address?.balance ?? 0)
   const { data: price } = useGetPriceQuery(currencies[currency].ticker, {
     pollingInterval: 60000,
     skip: totalAddressBalance === BigInt(0)


src/contexts/AnalyticsContext.tsx
```diff
@@ -74,34 +75,27 @@ const AnalyticsSetup = ({ children }: { children: JSX.Element }) => {
...
-  const analyticsId = useAppSelector((s) => s.settings.analyticsId)
-  const settingsLoadedFromStorage = useAppSelector((s) => s.settings.loadedFromStorage)
-
-  return (
-    <PostHogProvider
-      apiKey={PUBLIC_POSTHOG_KEY}
-      options={{
-        host: PUBLIC_POSTHOG_HOST,
-        disableGeoip: true,
-        enable: settingsLoadedFromStorage && analytics && !!analyticsId,
```
nop33, line 93: It could be that this line was the issue? :/

nop33: The docs say:

```
    // If set to false, tracking will be disabled until `optIn` is called
```

https://posthog.com/docs/libraries/react-native#configuration-options

src/components/AddressCard.tsx
```diff
@@ -59,7 +59,7 @@ const AddressCard = ({ style, addressHash, onSettingsPress }: AddressCardProps)
 
   const [loading, setLoading] = useState(false)
 
-  const totalAddressBalance = BigInt(address?.balance ?? 0) + BigInt(address?.lockedBalance ?? 0)
+  const totalAddressBalance = BigInt(address?.balance ?? 0)
```
nop33, line 62: I can't believe we did not notice this mistake... User reported it at #353

mvaivre: Wth!

mvaivre (approved):

